PR #151, opened 2018-10-17
Updated SQL Files

c11:
I updated the two sql file so that it can run successfully on my postgres database. If you merge, please only merge the two sql file.


snasaio:
This PR has been open, and is now out of date

PR #179, opened 2019-01-25
Refactor collection page to work with better with the new features

sirmspencer:
Common react improvements include
- Do not pass state into functions which can access the state
- Do not need state variables which are equivalent to other state variables + logic
- Do not name local variables the same as state variables
- Store state as close to children as possible
- Conditionally render as close to parent as possible

Move all calls to update map to be dependent on state by calling from componentDidUpdate

Reset to top question when visiting a new plot, new or analyzed

Do not highlight selected button

Update plot color based on selected question / highlight current question / update current current question on click, answer change, top level change

Show +/- for each question 

Refresh project stats each open instead of tracking locally (which includes only current user updates), hide on change plot

### DIFF
--- a/src/main/resources/public/jsx/collection.js
+++ b/src/main/resources/public/jsx/collection.js
@@ -664,11 +664,13 @@ class Collection extends React.Component {
     }
 
     render() {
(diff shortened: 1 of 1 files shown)


sirmspencer:
This PR is somewhat a WIP and is now superseded by the complete work.

PR #183, opened 2019-01-27
updates for local postgresql on linux and server

sirmspencer:
The server and windows seemed to do -host=localhost by default.  The param is needed explicitly for local Ubuntu (Any UNIX according to docs) servers.

Remove windows \r from .sh files

Explicit copy in data_migration.py to ensure script files exists (before it took a maven build to copy the script files)


sirmspencer:
merge approval from Mattermost

PR #189, opened 2019-02-03
Eliminate mutations pt 1 (with some bug fixes)

sirmspencer:
Common react improvements for `home` and `review-projects`
Remove use of `let` and `var` for `collection`, `home`, `review-projects`, and `mercator-openlayers` by refactoring mutating functions
Convert ajax to fetch on `review-projects` 

On the home page
- Fix bug where hide side panel button covers scroll bar
- Convert filter to immutable (using .filter) 
- Add sort options

In SurveyQuestions
- fix bug for survey card colors, which shows up with one survey card
- Update style for radio button, input box, and dropdown

### DIFF
--- a/src/main/resources/public/js/mercator-openlayers.js
+++ b/src/main/resources/public/js/mercator-openlayers.js
@@ -16,7 +16,7 @@
 ***
 *****************************************************************************/
 
-var mercator = {};
+let mercator = {};
 
 /*****************************************************************************
 ***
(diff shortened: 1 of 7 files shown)


sirmspencer:
All comments have been incorporate with the exception of a few:
create-institution and review-institution need a complete refactor, the comments will naturally be resolved during that process. (this is lower priority at the moment)
create-project also needs a final code cleaning, and replacing the ajax calls will be part of that. 

For replacing filter I used the following translations:
```filter()[0] -> find()
filter().length > 0 -> some()```

PR #191, opened 2019-02-04
Update survey design

sirmspencer:
Update survey design to use state instead of DOM access.  Make everything functional
Complete simple / advanced mode
- Prompt user if need to convert advanced to simple
Show more details for survey question (parent question, parent answer, type in advanced mode)
Correctly remove sub questions

Includes updates from comments for #187


sirmspencer:
All changes were made with the single exception `Line 80: This function seems to be popping up in several files. Refactor it into jsx/utils/SurveyUtils.js.`  I would like to group this into a larger task to refactor all common functions for the survey, convert to using index for the question and answer arrays, change to camel case, and update the function to convert old data.

PR #194, opened 2019-02-05
fixes from code review and more

sirmspencer:
Includes the fixes from review of PR #189 and PR #191 

Updates for the flow for plots with a single value
- Always act if the sample is selected (user does not need to ever select it)

Fix survey card sort order
Clear inputs after saving question (except parent question) and answer
Begin converting classes to use arrow function

### DIFF
--- a/src/main/resources/public/jsx/collection.js
+++ b/src/main/resources/public/jsx/collection.js
@@ -30,21 +30,6 @@ class Collection extends React.Component {
             selectedQuestionText: "",
             sampleOutlineBlack: true
         };
-        this.setBaseMapSource = this.setBaseMapSource.bind(this);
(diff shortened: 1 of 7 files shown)


sirmspencer:
* Line 591: Do you want to include [question] in the final result?
* Line 608: This is going to clear the toplevel question as well. Is this intentional?
--- You mentioned this before.  This is needed for the recursion to work as far as I can work it.  It is also OK for the original question to be include, as a new value for that question is being added. (before I filtered it out and you comment on that as well)

* Line 44: Change !every(!==) to every(===).
--- Changed to !some(!==), this will stop check once it finds a non button

* Line 140: Is componentTypes in scope?
--- componentTypes is a global constant

PR #197, opened 2019-02-06
fixes for review from 194

sirmspencer:
* Line 591: Do you want to include [question] in the final result?
* Line 608: This is going to clear the toplevel question as well. Is this intentional?
  --- You mentioned this before.  This is needed for the recursion to work as far as I can work it.  It is also OK for the original question to be include, as a new value for that question is being added. (before I filtered it out and you comment on that as well)
* Line 44: Change !every(!==) to every(===).
  --- Changed to !some(!==), this will stop check once it finds a non button
* Line 140: Is componentTypes in scope?
  --- componentTypes is a global constant

### DIFF
--- a/src/main/resources/public/jsx/collection.js
+++ b/src/main/resources/public/jsx/collection.js
@@ -57,7 +57,7 @@ class Collection extends React.Component {
             {
             this.showProjectPlots();
         }
-        // initiallize current imagery to project default
+        // initialize current imagery to project default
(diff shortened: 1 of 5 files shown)


lambdatronic:
Line 44: Change !every(!==) to every(===).
--- Changed to !some(!==), this will stop check once it finds a non button

Please use every(===) as I requested. every() also early exits when it find the first non-button, and a positive query is much easier to read and understand than a double negative.

sirmspencer:
I changed it to `every(===)`

PR #202, opened 2019-02-07
added dir variables in the sh scripts to avoid pathing issues

billyz313:
I added directory variables in the startceo.sh and the startgeegateway.sh to avoid any path issues.


sirmspencer:
Add a return character at the end of files for github.

PR #203, opened 2019-02-07
Githika ui home

gtondapu:
fixed scroll issue as i have added the show/hide filters. please verify this one and merge


sirmspencer:
283 `let filters = (props.showFilters) ?` use `const`

PR #204, opened 2019-02-07
Various UI fixes

sirmspencer:
Facing the user
- Prompt user on duplicate question and append enumerator 
-- Show in designer, hide elsewhere
- Validate project and survey before submitting.
-- Show selected file name on plot and sample design.
- Fix filter on home 

Behind the scenes
- Refactor convert survey values to survey quesions
-- Update to convert snake to camel
-- Add defaults for componentType and dataType
-- Force type for the few bad projects
-- Store result as `surveyQuestions` to distinguish good data and allow reference in the code

- Store surveyQuestion as selected instead of just text
-- Eliminates many conversions from text to id and back
-- Use .id everywhere that does not directly touch userSamples (still uses keys)

- Update create-projects page
-- Move map to life cylce
-- Use arrow functions
-- Update some of the state saving
-- Initialize new project in state instead of API call to empty project
-- Remove encode64() which was not doing anything, make plot and sample design stateless

### DIFF
--- a/src/main/java/org/openforis/ceo/utils/ProjectUtils.java
+++ b/src/main/java/org/openforis/ceo/utils/ProjectUtils.java
@@ -381,7 +381,7 @@ public class ProjectUtils {
 
     public static void runBashScriptForProject(int projectId, String plotsOrSamples, String script, String rpath) {
         try {
-            System.out.println("Runnin " + script);
+            System.out.println("Running " + script);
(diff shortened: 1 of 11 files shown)


lambdatronic:
Disregard my note on create-project.js: Line 150: "shp" should be "csv". I misread your code the first time through. It is already correct.

PR #205, opened 2019-02-09
Billyz geodashhelp localization

billyz313:
Spanish localization is added for the Geo-Dash Help Center


lambdatronic:
@billyz313 I corrected your code in Views.java. Please take a look at the geodashhelp() function in that file for an example of how to pass additional variables to your FTL files. Please do not modify the makeRoute() functions in that file.

PR #209, opened 2019-02-12
Fixes from review of 204

sirmspencer:
See #204 

Also includes file clean up per comments in the ceo slack channel.


lambdatronic:
Additionally, card-test.ftl and test-layout-editor.ftl should not be deleted unless their routes in Server.java are also removed. Have you checked with Billy to ensure that he no longer needs test-layout-editor.ftl?

sirmspencer:
I added back the seemingly unused flt files because they do still have a route.  Ill defer to Billy on those.
 
Added semi colons to all of the arrow functions.

`return new Promise.reject("Error creating project")` for failed fetch.

PR #210, opened 2019-02-12
Matt visual review

sirmspencer:
Move loading notifications on collection page under each sections header
Minor styling
Do not need to show archived date since users cannot access archived projects
AOI boundary numbers should be read only since they do not affect the map


sirmspencer:
`* SurveyCollection to SurveyQuestions incorrectly.` per the last PR it should be `SurveyCollection`.  I update all references in this PR just to be sure it is consistent.

PR #217, opened 2019-02-20
Clean and review institution pages

snasaio:
- Remove remaining unnecessary code in each review and create project, left from the split.
-- Update page routes
-- Update ftl
- Create a component for the common institution editor
- Update API to split create and update institution into individual calls
- Update API to split getAllUsers and getInstitution users into individial calls
- Use base64encoding for uploading file, eliminate form object.
- Update UI
-- Only show delete on imagery which is owned by institution
-- Show project privacy
-- Show project status with red/yellow/green
-- Hide email list for non users
-- Hide pending members for non admins
-- Show user role for members
-- Style information block


sirmspencer:
OfGroups.java:

* createInstitution() and updateInstitution() have not been updated to accept JSON data instead of MultipartForm objects.
--- Are we maintaining this database?  Im sure there are quite a few places where the functionality has not been updated.  I do not have a way to test any changes either.

collection.js:

* Line 27: Why did you remove saveValuesButtonEnabled?
--- This value is calculated and no saved in the state

create-institution.js:

* Line 39: Have you verified that this goes to the .catch() function when an error occurs?
--- Yes.  If we throw a reject, we also need a catch.  I also tested no reject, and it seems to be just as happy, nothing is passed to the next .then() if nothing is resolved. 


home.js:

* Line 558: Are you sure that the unmatched "i" tag is valid HTML?
---- any tag can be self closing in React.  However, I kind of glossed the whole "i" tag thing over, letting myself think i is image.  The i tag is probably not the most appropriate here in general, so I changed them all to span or div depending.

review-institution.js:
---  Sorry for making you review this, it looks like I missed a git push from my final review.


* Line 609: Need to add a test for the error condition first: data === "" ? "" : ...
---  My way was intentional so the error condition will default to red, along with any other abnormal situations.  With the opposite, abnormal would end up as yellow. the opposite logic would look like:
boxShadow: data === "" || data.analyzedPlots === 0
			? "0px 0px 8px 1px red inset"
			:  data.unanalyzedPlots === 0
				? "0px 0px 8px 1px green inset"
				: "0px 0px 8px 1px yellow inset"


* Line 750: Illogical test. Why would you test if this.props.userId === 1 here?
---- We have set up user 1 as the super user (admin@o----f---.org).  We also then hide user one from the users list API so we cant just check against the returned results

* Line 774: Does this override the sort on line 773?
---- This seems to correctly order by role then email


load_ceo_functions.sql:

* Line 1086: Is this change intended?
---- Yes, this now matches the JSON logic

sirmspencer:
* isInstitutionMember() test is still illogical. You need to compare findUserByEmail(userEmail).id === 1 NOT this.props.userId === 1
---- You are correct.  I miss understood what was wrong with the code.

sirmspencer:
* Line 777: Did you update SurveyCollection.js to take a selectedSampleId prop?
--- This is the merge error I referenced.  The prop is needed for highlighting answers.

PR #218, opened 2019-02-21
Extend base 64 uploads to create project, minor UI fixes

sirmspencer:
Updload csv and shp files using base 64 encoding

On the UI 
- fix sidebar scroll
- Add elipses to answer and question buttions
- Add title to all elipses to get hover over full answer.
- Correctly select card when question is changed to a different card
- Dynamically size project list on homepage

Fix issue with postgres and files not matching json names.

Dont allow user to delete an answer which is referenced by a sub question


sirmspencer:
ProjectUtils.java:

* Line 389: Why are we switching to /bin/bash? On most GNU/Linux systems, /bin/sh is a symlink to /bin/bash. On my machine, /bin/bash does not exist.
* Line 391: Why aren't we logging the output anymore? Is this not helpful?
---- For some reason I can use bash regex from the command line, even with sh, but not from the app.  I was able to replace with sed regex and its fine, we will have to have a windows user test the new way. The out.txt was a work around for windows I didnt need, but I supposed it can be helpful.

create-project.js:

* Lines 139-152: Fix indentation.
A single indention from fetch to then/catch matches the current eslint rule (and the rest of the project).  I did some research and it seems to be the more common format.  

home.js:

* Line 248: What happened to the height param? We introduced a complex ternary in PR217, but now it is completely gone. You mentioned making this a flex component but I don't see any flex styling. Is this cascading from your change on line 225?
--- That is correct, the parent div is now flex and has a height set.

PR #223, opened 2019-03-01
Add user locks to plots to prevent multiple users from analyzing the same plots

sirmspencer:
Split plot type functions from the project files

Lock plot for each un-analyzed plot requested by user. 
- Initial lock for 5 minutes
- The lock will continue to be renewed as long as the user remains on the analysis page
- Previous lock will be remove on any subsequent requests for un-analyzed plots
- Lock will be removed upon clicking quit, saving samples, or flagging a plot

Update fetch calls to use catch for more explicit error handling


sirmspencer:
create-project.js:

Line 249: Why are we no longer including this.state.projectDetails in the projectList?
--- I am confused by this comment.  This should be a 1 for 1 re-write.  Is it possible the diff is mixing with the getImageryList() function below

home.js:

Line 20: Include an alert "Error retrieving the project list. See console for details.".
Line 68: Include an alert "Error retrieving the imagery list. See console for details.".
Line 189: Include an alert "Error retrieving the institution list. See console for details.".
--- These are all started independently in different components.  A 404 error could cascade into the user getting 3 error messages. I changed them to use Promise.all like I did on the collection page to eliminate this issue.  There may be other locations like this.  *Post original comment: thinking about it, the likelihood of 3 404 errors but the page loading is probably nill. 

review-institution.js:

Line 31: Did you still mean to call this.setState({ projectList: [] })?
--- I do not see the same thing as you on line 31, a global search does not reveal `this.setState({ projectList: [] })`

review-project.js:

Line 104: Why did you stop setting availability to archived when archiving a project?
--- That was only setting the local state.  The user is immediately redirect and the state is dropped so there is no point in updating it.

PR #231, opened 2019-03-07
Githika surveyrules2

gtondapu:
Added survey rule design section with 4 rule types. It works for create-project now. Need to add it to collection and review-project once this works.


gtondapu:
I have retained my code for incompatible-answers block
Array.from is required as I tried other way and it throws error
updateQuestionId was not used anywhere, so I removed it. I get question ids from updateOptions function

PR #234, opened 2019-03-09
Optimize merge script

sirmspencer:
For the "old" method, use the faster json.load (vs demjson)
For the "new" method pass the json data as a string instead of using /copy inside PG
Script can resume easier.


sirmspencer:
Lines 594-598: Fix indentation.
---this is intentional logic for "resuming".  The scrip will only run adding projects.

PR #242, opened 2019-04-05
use quotes for json keys

sirmspencer:
The migration script json interpreter needs quotes for the json keys.


sirmspencer:
Merging this will cause maven to see these as newer than the local JSON database.  If this is merged, everyone will need to backup their database.  This is also more of a pain in Windows because that backup will maintain its modified date which is before the new files (ie there is no equivalent to the touch command)

sirmspencer:
I moved the files to test to avoid the problem above

PR #244, opened 2019-04-11
Matt test migration

sirmspencer:
Add password hashing
Remove whitespaces
Convert tabs to spaces
Use _uid and _rid
Restructure some functions
Clean up non conforming function formatting


sirmspencer:
load_ceo_functions.sql:

-Line 9, 52: plotId -> plot_id / sampleId -> sample_id.
--- this is correct, we denote the plot id in the file that can be any number as 'plotId' and 'sampleId'

-Line 23: as -> AS.
--- I use lowercase for column alias and uppercase for table and function assignments.

-Line 67: Why plotid but sampleId?
--- Typo, since SQL is not case sensitive it wasnt an issue.

PR #239, opened 2019-03-29
fixing rulesViolated

gtondapu:
1. added sample Id check for incompatible-types and sum-of-answers. sum-of-answers still needs some work. 
2. fixed create project bug: when a project is created from template, rules are not shown, so fixed it.

PR #246, opened 2019-04-16
rules load from template project

gtondapu:
Rules were not loading from template project, fixed it!


sirmspencer:
This should not be needed based on the change you already made in #239

gtondapu:
That is what I thought when I was looking for this change, so is that deployed?

PR #271, opened 2019-05-15
added geodash button

gtondapu:
Got to Geo-Dash button in collection in case user closes the tab


sirmspencer:
This has been incorporated into a larger PR to come.

PR #272, opened 2019-05-16
Update sidebar

sirmspencer:
-Open Geo-Dash button in collection in case user closes the tab 
-Add button to zoom back to plot if user zooms and pans map
-Add reset all survey answers by save button 
-Move go to plot feature to a less UI forward location (many users clicked it not knowing what it’s for and since it’s a feature only a few would use it should not be the top item)
-Move next, skip, back buttons to bottom by save buttons ?
-Rethink colors (Not started, progress, complete) to be 502 compliant on surveys (red or green very bad for color challenged people
-Make buttons bold with a thicker boarder so they have higher contrast.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -74,6 +74,7 @@ module.exports = {
         "no-console": 0,
         "no-multi-spaces": 2,
         "no-useless-return": 1,
+        "no-unused-vars": 1,
         "no-var": 2,
         "object-curly-newline": [1, { "multiline": true, "consistent": true }],
         "object-curly-spacing": [1, "always", { "objectsInObjects": false }],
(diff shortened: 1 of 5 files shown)


sirmspencer:
collection.js:
I just noticed that the project title bar had never been updated to show the ellipses for too long of a name.  I added a quick update for that. 

SurveyCollection.js:

Line 175: Why is the green box-shadow 5px while the yellow and red ones remain at 4px? This is inconsistent with line 64 above.
review-institution.js:

Line 594: Why is the green box-shadow 2px while the yellow and red ones remain at 1px?

---- The new blue color is lighter than the other two so I made the shadow a little thicker.  However the survey card button did not need the extra width.

PR #286, opened 2019-05-17
Billy add asset to basemap

billyz313:


sirmspencer:
collection.js
160-162 -> this.setState({hasGeoDash: widgets.length > 0})
165-167 You can remove this.  You do not need to handle a reject just to throw it again.

geo-dash-widget-editor.js
285 This should be an alert. Also console log the responce.
673 remove space before :
1177-1188 (x5) This is repeated code can you create a class function to return this checkbox
1185 This code has access to the state direction, you dont necessarily need to create a function to update the state. If you want you can use `onChange={() => this.setState({ addCustomImagery: event.target.checked })}`

PR #289, opened 2019-05-20
template bug

gtondapu:
When you go to create project page and select a template project, it populates the fields. But, when you go back and select "None" in template project list. It does not take away the fields and bounding box. Fixed this!


sirmspencer:
create-project.js
232-251 These are the same values we initialize the the project with on load.  Since the are the same in two places, create a const blankProject = {} at the top (above `class Project`) to store these values

If you are resetting everything, the coordinates also need to be set back to ""

255 Please add the trailing comma back. One of the linter rules I added requires a comma on the last item of the list.  This help keep diffs in Git a bit simpler because adding a new item does not require a new comma on the preceding line.  This means the diffs will only show the new line.

PR #293, opened 2019-05-21
getting imagery from institution imagery now

billyz313:


sirmspencer:
geodashreact.js
44-65 Your two calls can be done in parallel.  If you want two seperate messages keep them seperate using two functions like getImageryList called from component did mount with their own catch.  Otherwise use promise.all like in home.js

445-454
```getInstitutionBaseMap = (basemap) => !basemap 
		? this.props.imageryList[0]
		: this.props.imageryList.find(imagery => imagery.id === basemap.id);```

or remove and replace 457 with 
```const raster = this.getRasterByBasemapConfig(!widget.baseMap
    ? this.props.imageryList[0]
    : this.props.imageryList.find(imagery => imagery.id === widget.baseMap.id));```


792, you cannot set the state directly

sirmspencer:
- the code u want me to replace 445-454 with does not do what it's supposed to do, maybe i'm not understanding what ur going for there...
   - dont worry about it.  your use of a function instead of mutating a variable is already looking good.

- also 44-65 Cannot be done in parallel, they must be done in that order get-all-imagery, then get widgets.  The promise.all gets them in parallel so the widgets states get set before the imagery and then fail to render since they need this information.
 - why not? I do not see any information from one that is used in the other.  If you need both pieces to render, you can use conditional rendering elsewhere so that the data can be fetched in parallel.

billyz313:
- why not? I do not see any information from one that is used in the other. If you need both pieces to render, you can use conditional rendering elsewhere so that the data can be fetched in parallel.
For some reason the widgets get back before the imagery and when they try to render they crash because there's no imagery in the imagery list.  after this I can spend a little time to add conditional rendering (if no imagery) don't render widgets and see how that plays out.

PR #297, opened 2019-05-23
more linting

billyz313:


billyz313:
I made all the changes with the exception of the Lines 237-242 which does not evaluate the null properly in the style {{  }} that way.

PR #299, opened 2019-05-24
Matt cleanup

sirmspencer:
Remove unused files
Update file names to match routes
Convert some font-awesome icons to unicode.
Clean up some notes
Enable cache
Update webpack settings so it is (hopefully) the same in dev and prod.


sirmspencer:
--remove custom.js script tag
----As far as I can tell this is still being used.

--Line 990: Use UTF-16 encoding, not UTF-32. JS strings are always in UTF-16.
----I switched to html format (see the full font awesome pr).  Its easier to look up than the 2 character UTF-16 version

need to delete these files:
--test-jerome.ftl 
----I think it already is. 
--fontawesome-all.min.js is still being used in this PR.  
----The full replacement is in a separate PR.

--glyphicons-halflings is still used in geo-dash.css (can this be switched out?)
----Im not sure.  I still have not really dug into the geo-dash.

PR #302, opened 2019-05-28
WIP Auto generate script tags for freemarker templates

sirmspencer:
This was developed on Linux but should work on Windows using git bash.


sirmspencer:
This branch has a messy commit history.  I am going to clean that up and re-do the sh script in JS.

PR #304, opened 2019-05-29
WIP Convert highcharts and jquery to npm

sirmspencer:


sirmspencer:
This will be easier if we convert the date picker to react first.  Ill cherry pick on top of that change.

PR #311, opened 2019-06-11
Move institution page project status to mouse over

sirmspencer:


sirmspencer:
Now that we have seen another cause for the massive slowness on the server, this is likely not needed.

PR #321, opened 2019-06-20
Replace spinner with modal

sirmspencer:
Remove utils.js
Remove announcement


sirmspencer:
--Line 148: const isNumeric = n => !isNaN(parseInt(n)) && isFinite(parseInt(n));
----From the docs for isFinite (not Number.isFinite): `If  needed, the parameter is first converted to a number.`.  It is almost sufficient on its own, with the only exception being whitespace.  I suppose `!isNaN(parseInt(n))` ends up being a hack way to check for white space strings. `n.trim().length === n.length` would be more clear.

PR #322, opened 2019-06-21
Allow users to add bing and planet imagery

sirmspencer:


sirmspencer:
-Line 450: Add a space before the closing }.
----There is no space because it is also a } ( }} ok, } } not).  Its built in to the rule and auto corrects to how it is.

PR #326, opened 2019-06-27
Create archiving and restoration scripts

sirmspencer:
Use external settings.conf to enable archiving
Script to set up file and folders correctly
Use permissions so openforis user can download, script included to download and restore to local computer
Script to restore DB from local backup.  The recovery.conf file can be modified to allow point in time recovery.


sirmspencer:
The logic allows us to delete data, then archive OR close. But only close if it's. It already archived. I added this for the clean up plan. For some of the projects with too many plots, the user may want to reuse the survey data.

lambdatronic:
```sql
IF _archive THEN
￼        PERFORM archive_project(_project_uid);
￼    ELSE
￼        PERFORM CASE WHEN availability <> 'archived' THEN (SELECT close_project(_project_uid)) END
￼        FROM projects
￼        WHERE project_uid = _project_uid;
￼    END IF;
```
When I read this conditional, I see this:

1. If we passed _archive = true (the default in PostgresProjects.java), then we set the project's availability to "archived".

2. If we passed _archive = false (does this ever happen?) and the specified project's availability != "archived", then we set the project's availability to "closed".

3. If we passed _archive = false and the specified project's availability == "archived", we do nothing.

Here are my questions:

1. Why are we closing projects in the delete_project() function? This is extremely unintuitive hidden behavior. So now delete_project(id, true) means archive and delete_project(id, false) means close. This kind of conditional logic should be done on the client side, so we can call delete_project(id) or close_project(id) appropriately.

2. Even if you did want to implement something this confusing, why would you put the conditional test on availability in a CASE WHEN expression within a PERFORM within delete_project()? You should simply call PERFORM close_project(_project_uid) and be done with it. If you want to prevent archived projects from having their availability set to "closed", you should put that conditional logic in the close_project() function as it is a check that applies universally.

sirmspencer:
I reverted the SQL changes.  I can agree that these changes dont make sense in the frame that they are only used for my data clean up plan.  We still have that to go over, and I can later make a separate temporary function that has a more clear name and logic.

PR #329, opened 2019-07-03
Better default filter for homepage institutions

sirmspencer:
Also change default privacy level to institution.


sirmspencer:
Line 72: SideBar doesn't use userId prop.
--- It does (filterHasProj()), but I can eliminate userName

PR #332, opened 2019-07-04
Remove un-maintained collect DB

sirmspencer:


sirmspencer:
web.xml:
Add databaseType parameter.
CeoSparkFilter.java:
Add databaseType parameter.
CeoConfig.java:
Add databaseType parameter.
Server.java:
Make init() call declareRoutes() for either the JSON or POSTGRES DBs based on CeoConfig.databaseType.
---I see the additional data to remove, but why do we need a CeoConfig.databaseType?  It would not be used anywhere.

CEO_Developer_Documentation.txt:
Lines 7-14: None of this documentation makes sense without the Collect and OfUsers bridge code in CEO and should be removed/rewritten.
README.org:
Update the documentation to describe the current steps to launching CEO using Maven, Gradle, or Tomcat.
---I am not familiar with Tomcat or Gradle at all.  If the Tomcat/Gradle portion needs to be updated, can you handle this?

lambdatronic:
- ProjectStats.java:
  - Delete this file. (It was only used by CollectProjects.java.)
- web.xml:
  - Re-add baseUrl and smtpUser parameters.

PR #334, opened 2019-07-05
Delete projects with delete institution

sirmspencer:
There is no way to access a project if its institution is deleted, so delete it also.


sirmspencer:
Shouldn't we also be setting availability = "archived"? Do we still have an archived field in this table?
--- The only column is "archived" for institutions.

PR #340, opened 2019-07-17
3 bug fixes

sirmspencer:
1. Use Promise.all instead of passing a promise to Promise.reject()
2. Reset template params when toggling "Use Template Plot"
3. a. Update check for simple/advanced mode to properly fire on changes
    b. Check for rules before switching to simple mode
    c. Clear rules when switching to simple mode


sirmspencer:
SurveyDesign.js:
Line 27: My understanding was that inSimpleMode: props.hasNoRules was there to ensure that projects with rules (on the Review Project page) would display automatically in advanced mode. Why are you setting this to true?
---This is only the default, the state changes along with the template (line 37).  Githika's code to rely on props was flawed because the state is only set = props.hasNoRules on mount.  This will always be false because this component is mounted when the page loads initially with no template. Setting the state to a prop is a red flag.  I will always do it in didMount to make it obvious its intentional and that I understand it will only happen the one time.


create-project.js:
Same issue as in SurveyDesign.js: Why are you eliminating the hasNoRules prop? This is to ensure that template projects which contain rules will automatically be displayed in advanced mode.
---This logic was flawed.  I am also handling the change inside the SurveyDesign component where it should (IMO) be handled. 
Lines 280-282: What is the purpose of this code? It looks like you are preventing the user from accessing the templateProject's surveyQuestions, surveyRules, and privacyLevel.
---The check box that is triggering this function is "Use Template Plots".  I am reverting the only the template related to plots so the other user changes stay.  For example the user unchecks "Use Template Plots" and makes a change to the number of random plots, when the box is checked again, we want to use the number of random plots from the template again.  On the back end with the box check, it really ignores the input.  By reverting the changes, the user sees the correct values for the new project in the read only plot information display.

PR #341, opened 2019-07-17
Proxy GeoServer imagery

sirmspencer:


sirmspencer:
Line 47: It would be good to bind geoserverParams.keySet() to a variable on the next line as it will be used repeatedly in the following stream operation:
-- The second use should have been entrySet, so there is no longer repeated use.

Line 51-52: Performing the split twice doesn't seem particularly efficient. Better to bind the result to a variable:
--  It looked prettier this way / not and issue on modern processors.  But not enough to argue.

PR #344, opened 2019-07-23
Billy convert ghc to react

billyz313:
Converted Geo-Dash Help Center to react and redesigned the format of the UI


src/main/resources/public/jsx/geo-dash-help.js
```diff
@@ -0,0 +1,357 @@
...
+                        rel="noopener noreferrer"
+                    >OpenForis-SEPAL</a> <span>{this.state.LngObject.post_sepal_link}</span>
+                </p>
+            </div>
+        );
+    }
+
+    getLi = text => <li style={{overflowWrap: "break-word"}}> {text} </li>;
+
+    toggleHelpImage = event => {
```
sirmspencer: Accessing the DOM / document directly is not  the preferred way in react if we can help it.  This page is not so react heavy that it would matter much, but this is a great example of re-usable code.  If you can convert this into a stateful component <GrowableImage> we can use it on other help pages.  Shorthand code:

```
class GrowableImage
state={fullSize: false}
const imgStyle = this.state.fullSize ? {full size style} : {normal size style}
<img style={imgStyle} onClick={this.setState{ fullSize: !this.state.fullSize />
```

src/main/resources/public/jsx/geo-dash-help.js
```diff
@@ -0,0 +1,357 @@
...
+                                                <span data-tag="enter_image_asset_params"> {this.state.LngObject.enter_image_asset_params}</span>
+                                                {"{\"bands\":\"B4,B5,B3\",\"min\":\"10,0,10\",\"max\":\"120,90,70\"}"}
+                                            </li>
+                                            {this.getLi(this.state.LngObject.skip_date)}
+                                        </ol>
+                                    </li>
+                                    <li>
+                                        <span data-tag="image_collection_asset">{this.state.LngObject.image_collection_asset}</span>
+                                        <ol style={{listStyleType: "lower-roman"}}>
+                                            <li style={{overflowWrap: "break-word"}}>
```
sirmspencer: This is redundant with getLi()

src/main/resources/public/jsx/geo-dash-help.js
```diff
@@ -0,0 +1,357 @@
...
+                                                {this.getLi("Offline Carbon Monoxide - COPERNICUS/S5P/OFFL/L3_CO {\"bands\":\"CO_column_number_density,H2O_column_number_density,cloud_height\",\"min\":\"0\",\"max\":\"0.5\"} \n   2018-01-01 to 2018-12-31")}
```
sirmspencer: Should this be in the language file?

sirmspencer:
Final comments 
- in your language files select_image_collection is a duplicate key.
- You do not need createListItem (fka getLi).  You can set overlowWrap at the parent div. 
- You dont need <span> around {this.state.etc}
- `data-tag` doesnt seem to do anything, I think you could remove these spans as well.

lambdatronic:
- FormComponents.js:
  - Line 38: componentDidUpdate() should be defined as a method rather than as an anonymous function.
  - Lines 136-154,159-172: These are stateless ifs. Convert to "return ?:".
  - Line 155,173: Missing semicolon.
- geo-dash-help.ftl:
  - Line 9: `section id="geo-dash-help"` <-- id should correspond to the page name
  - Lines 13-14: Remove excessive space between attributes and values.
  - Line 18: `<div id="dashHolder"></div>` should be placed directly before the script tag to ensure that it exists before the JS code is run.
  - Line 19: logo-banner.ftl's #include should be moved directly above end-content.ftl's #include.
  - Line 20: Remove dangling `</div>`.
- geo-dash-help.js:
  - Line 9,*: Incorrect attribute camelCasing: LngObject -> lngObject
  - Line 13: componentDidMount() should be defined as a method rather than as an anonymous function.
  - Lines 19,29: No enclosing parentheses.
  - Line 28: Missing semicolon.
  - Lines 61,62: Excessive spaces are auto-eliminated in HTML. Why are these still in the code?
  - Line 198: Combine the two strings with "+" as on line 168.

PR #348, opened 2019-08-02
Fix proxying of Planet layers

lambdatronic:


sirmspencer:
collection.js
621 - ternary with objects are hard.  As an alternate fomrat: I did it a different way elsewhere is CEO, the way below shows the conditional on one line and then the data returned nested inside. Chose your favorite.

(this.state.currentImagery.title === "DigitalGlobeWMSImagery") ? {  <-- conditional
    imageryYearDG:     this.state.imageryYearDG,                    <-- nested data.
    stackingProfileDG: this.state.stackingProfileDG,
} : (this.state.currentImagery.sourceConfig.type === "Planet") ? {
    imageryMonthPlanet: this.state.imageryMonthPlanet,
    imageryYearPlanet:  this.state.imageryYearPlanet,
} : {};

1365 - use `new Date().getFullYear()` so we dont have to update every year.

PR #371, opened 2019-12-10
sepal-ceo integration

rfontanarosa:
- DB: a "token_key" field has been added into "projects" table
- FRONTEND: the create-project script now CONSUMES a JSON which includes the token key
- BACKEND: the token key is generated at creation time, the create-project endpoint now RETURNS a JSON which includes the token key, the checkAuthCommon method now checks if there is a token in the query string but only if the user isn't logged and allows the access of the collection page only if the token key is valid

### DIFF
--- a/src/main/java/org/openforis/ceo/local/JsonProjects.java
+++ b/src/main/java/org/openforis/ceo/local/JsonProjects.java
@@ -2,6 +2,7 @@ package org.openforis.ceo.local;
 
 import static javax.servlet.http.HttpServletResponse.SC_NO_CONTENT;
 import static org.openforis.ceo.local.JsonImagery.getImageryTitle;
+import static org.openforis.ceo.utils.DatabaseUtils.connect;
 import static org.openforis.ceo.utils.JsonUtils.elementToArray;
 import static org.openforis.ceo.utils.JsonUtils.elementToObject;
 import static org.openforis.ceo.utils.JsonUtils.expandResourcePath;
@@ -50,6 +51,7 @@ import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.nio.file.Files;
 import java.nio.file.StandardCopyOption;
+import java.sql.SQLException;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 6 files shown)


lambdatronic:
PostgresProjects.java:
- Lines 60,76: Use camelCase for the "tokenkey" session attribute for consistency.
- Lines 91,93: Fix indentation.
- Line 625: Remove extra whitespace at end of line.

JsonProjects.java:
- NEED: Repeat the same tokenKey logic as in PostgresProjects.java.

upgrade-2019-12-08.sql:
- Line 4: Add `DROP FUNCTION create_project(....);` before the `CREATE OR REPLACE FUNCTION` statement. This is necessary because you are changing the arguments to the function.
- Lines 55,57: No need to drop and re-add the editable attribute.

lambdatronic:
Just reposting these comments in case you missed them the first time...

PostgresProjects.java:
- Lines 60,76: Use camelCase for the "tokenkey" session attribute for consistency.
- Lines 91,93: Fix indentation.
- Line 625: Remove extra whitespace at end of line.

JsonProjects.java:
- Repeat the same tokenKey logic as in PostgresProjects.java.

upgrade-2019-12-08.sql:
- Add `DROP FUNCTION create_project(....);` before `CREATE OR REPLACE FUNCTION` because you changed the function arguments.
- Lines 55,57: No need to drop and re-add the editable attribute.

rfontanarosa:
Argh! I've missed it. Ok, I'll do these changes...

rfontanarosa:
hi gay, I've included all your suggestions except one and I've done more tests and fixed some stuff.

I can't fix "Lines 55,57: No need to drop and re-add the editable attribute." because editable must be the last attribute of project_return due other queries that adds the editale field at the end of the selections and there is no way to modify a type adding an attribute in a specific position:

-- Returns a row in projects by id
CREATE OR REPLACE FUNCTION select_project(_project_uid integer)
 RETURNS setOf **project_return** AS $$

    SELECT *, FALSE as editable
    FROM project_boundary
    WHERE project_uid = _project_uid

$$ LANGUAGE SQL;

lambdatronic:
JsonProjects.java:
- Line 5: DatabaseUtils.connect is never used.
- Line 54: SQLException is never used.
- Line 97: Remove trailing whitespace.
- Line 1335: Remove trailing whitespace.

PostgresProjects.java:
- Lines 91,927: Fix indentation.

create-project.js:
- Lines 148,149,152: You reverted your original change here. `data` is an object, so you shouldn't be accessing it like an array.

rfontanarosa:
Done, but... I remember what I did with the data object but today, when I did some tests after the creation of a project the page didn't  changed and it returned an error. After I debug I discovered from the console that data was an array.

PR #387, opened 2020-02-05
update tsServer and enable 15 minutes caching.

c11:


src/main/resources/public/js/ts_scripts.js
```diff
@@ -2743,13 +2745,11 @@ function appendChips(window, selected, color) { //this function is handling the
...
-    //TODO: restore these
-    //YANG: temporarily disable caching
-    // if (chipInfo && new Date(chipInfo.lastGatewayUpdate) > currentDate) {
-    //     return chipInfo.chip_url;
-    // }
+    currentDate.setTime(currentDate.getTime() - 10 * 60 * 1000);
+    const info = JSON.parse(localStorage.getItem(chip_url));
+    if (info && new Date(info.lastGatewayUpdate) > currentDate) {
+        return info;
+    }
```
sirmspencer: use explicit else { return ''; }

PR #397, opened 2020-02-14
add the end date for date range in planet daily

biplovbhandari:


lambdatronic:
mercator-openlayers.js:
- Line 180: "Error loading Planet Daily imagery. Check console for details."

review-institution.js:
- Line 451: Add endYear, endMonth, endDay params and split `const date` into `const startDate` and `const endDate`.
- Line 556: Check for endYear, endMonth, endDay params
- Lines 558-562: Re-align ?s
- Line 562: Check for startDate and endDate instead of just date

collection.js:
- Line 259: Add a period.
- Line 260: Use `else` instead of `return`.
- Line 270: Combine setImageryEndDatePlanetDaily with setImageryStartDatePlanetDaily
- Line 326: No parens around this.addDaysToDate. However, you should really just use imageryEndDatePlanetDaily (which should be specified on the Review Institution page as per my comment above).
- Lines 1511,1523: Two HTML tags can't have the same id.
- Lines 1514,1526: Use `min="2010-01-01"`

lambdatronic:
collection.js:
- Line 257: Rewrite in a functional style.
- Line 268: Include date order test as above.

review-institution.js:
- Line 574: "start" -> "end" after ||
- Lines 578,580: Remove extra set of parens.

Issue #403, opened 2020-02-20
Numerous fixes to SecureWatch imagery


lambdatronic:
1. Add a new type, called "SecureWatch", to the Add Imagery component
   on the Review Institution page. The user should input a Title,
   CONNECTID, Start Date, and End Date. Use date selector components
   for this, just like with Planet Daily imagery.

2. The SecureWatch imagery type should create a new entry in the
   database with these parameters.

   "title": "TAKEN FROM USER",
   "attribution": "SecureWatch Imagery | © Maxar Technologies Inc.",
   "sourceConfig": {"type":            "SecureWatch",
                    "startDate":       "TAKEN FROM USER",
                    "endDate":         "TAKEN FROM USER",
                    "geoserverUrl":    "https://securewatch.digitalglobe.com/mapservice/wmsaccess",
                    "geoserverParams": {"VERSION": "1.1.1",
                                        "STYLES": "",
                                        "LAYERS": "DigitalGlobe:Imagery",
                                        "CONNECTID": "TAKEN FROM USER"}}

3. When SecureWatch imagery is selected on the Collection page,
   provide two date selector widgets for the start date and end date
   as with Planet Daily imagery. Initialize these to whatever values
   comes from its sourceConfig.

4. On the Collection page, re-use the functionality from
   DigitalGlobeWMSImagery to set the COVERAGE_CQL_FILTER query
   parameter to
   "(acquisitionDate>='startDate')AND(acquisitionDate<='endDate')".
   Update these parameter with the new dates each time startDate or
   endDate are changed.

biplovbhandari:
closed by above PR

PR #405, opened 2020-02-23
fix to issue #403

biplovbhandari:


lambdatronic:
Issues:

1. SecureWatch imagery should appear in the Collection page's imagery dropdown even on the project overview screen.
2. Attribution dates are not set correctly when a plot is first loaded and/or SecureWatch imagery is first selected.
3. The CONNECTID is stored multiple times within the sourceConfig object. Store it only under geoserverParams.
4. The CONNECTID is not filtered out of the sourceConfig object before returning it when /get-all-imagery is called. This is a security risk.

Issue #402, opened 2020-02-20
Numerous fixes to Planet Daily imagery


lambdatronic:
1. Change Add Imagery component on Review Institution page to get the
   start date and end date from the user using two date components as
   on the Collection page rather than the 6 number inputs that are
   currently shown.

2. Use the map's view extent rather than the plot boundary as the
   geometry sent to Billy's Planet Daily proxy route.

3. Provide a default start date and end date on the Collection page
   for Planet Daily imagery in case they were not set when the imagery
   source was created.

4. Replace the LayerGroup with individual layers being added to the
   map, so that they can be turned on and off with a layer switcher
   control. Consider using a unique ID field in the layers' properties
   so that they can be removed together easily.

5. Add a layer switcher control to mercator.createMap.

PR #409, opened 2020-02-26
fix to issue #402

biplovbhandari:
fix to issue #402

### DIFF
--- a/src/main/resources/public/css/custom.css
+++ b/src/main/resources/public/css/custom.css
@@ -629,3 +629,34 @@ ul.samplevalue{
         width:100% !important;
     }
 }
+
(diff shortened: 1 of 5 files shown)


lambdatronic:
mercator-openlayers.js:
- Line 93: Remove `export` attribute and add PlanetLayerSwitcher to export list at bottom of file.

collection.js:
- Lines 7,392: Don't import fromExtent from openlayers directly into CEO. Map operations should be abstracted behind mercator so as to remain mapping library independent.

review-institution.js:
- Line 561: What type are startDate and endDate? Do we need to call `new Date()` on them before using `>`?

PR #410, opened 2020-02-26
Add mailing list functionality

### LINKED ISSUE #386: Add mailing list functionality
Add a new page that is only accessible to admin@openforis.org.

On this page, there should be two form fields (Subject and Body) as
well as a button labeled "Send to All CEO Users".

Pushing this button should send a mass email to all registered users
(i.e., everyone with an email address in the `users` table).

rfontanarosa:
Fixes #386 

The page **/mailing-list** is accessible only by the user who has its id equals to 1.

I've used the same style of the Account page and I've added CKEditor as framework to create a WYSWYG editor.

I've improved the Mail class, is now possible sending mail in text/plain and text/html format to a list of recipients, a list of cc recipients and a list of bcc recipients.


lambdatronic:
JsonUsers.java:
- Lines 544-545: Remove extra spaces after =.
- Line 547: inputSubject.isEmpty() -> inputBody.isEmpty()
- Line 552: Use var.
- Line 554: "Your message has been sent to the mailing list."
- Line 557: "There was an issue sending *to* the mailing list. Please check the server logs."

PostgresUsers.java:
- Lines 411-412: Remove extra spaces after =.
- Line 414: inputSubject.isEmpty() -> inputBody.isEmpty()
- Lines 418-431: Fix indentation.
- Line 420: Use var.
- Line 426: "Your message has been sent to the mailing list."
- Line 430: "There was an issue sending *to* the mailing list. Please check the server logs."

Mail.java:
- Line 7: Remove blank line.
- Lines 33-42: Eliminate intermediate variables and mutation with this version:
```java
return listString.stream()
    .map(email -> {
        try {
            return new InternetAddress(email);
        } catch (AddressException e) {
            //e.printStackTrace();
            return null;
        }
    })
    .collect(Collectors.toList())
    .toArray(new Address[0]);
```

mailing-list.ftl:
- Lines 5-44: This should all be done with React and not jQuery.
  Except for Yang's in-progress TimeSync translation, we haven't used
  jQuery in CEO for several years. The ClassicEditor call should be
  added to a componentDidMount lifecycle hook in mailing-list.js and
  the HTML and onClick event should be added to a render() method.
- Line 32: "#editor"
- Lines 36,37: Use ES6 arrow syntax for functions.
- Line 38: "Are you sure you want to send *to* this mailing list?"

sirmspencer:
This implementation will cause significant issues with our smtp server.  The total amount of users in the CEO database exceeds the limit for a basic smtp server.  Also you do not have anyway to unsubscribe, or track inactive users.  For a feature like this, you will need a service like MailChimp.

Without a third party service you should limit this to institutions sending to only their members, and consider making them use their own smtp server.

rfontanarosa:
> This implementation will cause significant issues with our smtp server. The total amount of users in the CEO database exceeds the limit for a basic smtp server. Also you do not have anyway to unsubscribe, or track inactive users. For a feature like this, you will need a service like MailChimp.
> 
> Without a third party service you should limit this to institutions sending to only their members, and consider making them use their own smtp server.

Ouch! We can use a service like MailChimp (good alternative to SendGrid, a service that I used in another project) but we have to consider their costs or if their free plan fits with our needs. Moreover we have to share with these services the emails of CEO subscribers.

Maybe we can use the internal functionality to send messages to a single user or to all the members of an institution and delegate the global mailing list to a service.

Anyway... how many users does CEO have?

PR #415, opened 2020-03-03
break up timesync UI into react component.

c11:
Gary, please check the coding style of the react components for TimeSync. The logics were not put in the component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     },
     "private": true,
     "dependencies": {
+        "d3": "^3.5.17",
         "lodash": "^4.17.15",
         "ogr2ogr": "^1.4.0",
         "ol": "^5.3.3",
--- a/src/main/resources/public/jsx/timesync.js
+++ b/src/main/resources/public/jsx/timesync.js
@@ -897,7 +897,3 @@ export function renderTimeSyncPage(args) {
     );
 }
 
-export {
-    ChipViewer,
-    TrajectoryPanel,
-};
--- a/src/main/resources/public/jsx/timesync.js
+++ b/src/main/resources/public/jsx/timesync.js
@@ -1,5 +1,5 @@
-import React from "react";
-import ReactDOM from "react-dom";
+import React from 'react';
+import ReactDOM from 'react-dom';
 
 class TimeSync extends React.Component {
     constructor(props) {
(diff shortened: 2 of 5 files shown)


lambdatronic:
timesync.ftl:
- Lines 2,28: <#-- is not a valid HTML or FTL tag. Use  for HTML comments.
- Lines 12-14: Do not set global variables in FTL files. Pass your parameters to timesync.renderTimeSyncPage and pass them from there to ts_scripts.js.
- Line 14: Don't repeat a computation that has already been bound to a variable:
```js
console.log()window.tsDashMessage;
```
- Line 25: Inline the timesync-script.ftl contents into timesync.ftl and delete timesync-script.ftl.
- Line 27: Remove dangling blank line.

timesync-script.ftl:
- `<script></script>` tags should be moved to header.ftl. However, whenever
  possible it is preferred that libraries are loaded through
  package.json and imported directly into your JS files rather than
  installed globally with <script> tags.

timesync.js:
- Please install eslint in your IDE. I have included a .eslintrc.js JS linting file that will automatically highlight any style errors in your files.
- Use "" not ''.
- Lines 8-9: Don't store props in state. Simple use props.documentRoot directly.
- Lines 9,572: userId does not appear to be used.
- Lines 12,33,40,80,82,99,100,101: Remove unnecessary blank lines.
- Lines 43,50,58: Be consistent and only use one blank line between your commented sections.
- Line 103: Remove development logs in production.
- Line 107: this.state.documentRoot -> props.documentRoot
- Line 108: Use this instead:
```js
.then(response => response.ok ? response.text() : Promise.reject(response))
.then(data => {
    console.log(data); // FIXME: Remove this in production
    this.setState({version: data});
})
.catch(response => {
    console.log(response);
    alert("Error retrieving the TimeSync info. See console for details.");
});
```
- Lines 126,128,129,131,133: Remove unnecessary blank lines.
- Line 130: Q: where should the following code go? A: Just make another component.
- Lines 147,241,253,264,276,346,545,553: When class components only contain a render() method, just make these into function components instead.
- Line 258: Missing closing `</ul>` tag.

c11:
Fixed code to addressed review comments:

1. <#-- is valid FTL tag.
2. the class components are placeholders and some of them might be converted to function components next.

sirmspencer:
timesync.ftl:
18,19, 20 It doesnt look like you are actually using these.  Remove these import.  If you chose to use one later use the npm package.  As a general concept, with React you should be able to avoid jquery and jquery based packages. I can certainly help if you like help avoiding jquery.
22 We already have lodash as a dependency in package.json. See widget-layout-editor which also imports lodash. (it also doesnt look like you are using it)
21, 23, We already have these in the public JS folder, import them directly in JS. See collection.js which imports functions from openlayers-mercator.js 

Please install eslint in your IDE. We have included a .eslintrc.js JS linting file that will automatically highlight any style errors in your files.

sirmspencer:
For your next branch can you change your naming a little bit? The branch starting with fb is the feature branch.  Each change you make is going into the feature branch but is not a feature branch.  Start you branch name with `yang` and then a short description of what you are doing.  For this branch something like `yang-initialize-react-page` would be more clear.

sirmspencer:
src/main/resources/public/jsx/timesync.js
- 775, 779, , 883, 882, 883 do not change the variable name as you pass it.  Update these `chipProps` to `chipSettings` as well.

PR #434, opened 2020-03-23
Upload csv: allow more columns than just lon, lat, plotid

rfontanarosa:


sirmspencer:
There does not need to be anychanges to this file if you use SQL and update select_partial_table_by_name to dynamically build a query with the extra columns.

Starting at line 52, select_partial_table_by_name, if you can make this query damically select all the remaining columns into extra_feilds, it will work for shape files as well.
Line 1500, there are no matching updates for this feild to be available.  I do not think you this information in the whole plot list.
Passing the extra fields should be done in select_next_user_plot, select_prev_user_plot, select_next_unassigned_plot, select_prev_unassigned_plot

Line 116, 322, these are not used

src/main/resources/sql/update_2020-03-20.sql
Lines 34-61, this is a duplicate definition
Line 65+ select_all_project_plots, I do not think you need to add extra_fields to the whole plot list.

sirmspencer:
Use `select_json_table_by_name` to get the the entire row as json, then you can filter the rem_data json object in Java to remove unnecessary colums (gid, plotId, center, geom)

lambdatronic:
Two comments:

- Make sure that a project manager can restrict which fields are displayed on the Collection page
- Will this update display extra fields from CSV projects that are already in the DB?

rfontanarosa:
I've just created a new branch (rfontanarosa_allow-more-csv-columns) based on master.

I've pushed only the sql script which changes the return type of the function listed by Matt above and added the joins to return the extra plot data and the extra sample data: https://github.com/openforis/collect-earth-online/commit/703d70e14d6037f4c8f986d4ce1ee5cb363e60e1

At the moment the interface shows the extra plot information which are stored inside the ext_ plot table as the previous branch.

I've added the buildPlotCollectionJson method. For the moment it doesn't filter the extra plot information, so it returns lat, lot, plotid too.

PR #436, opened 2020-03-31
aoi error using project template

### LINKED ISSUE #435: creation of non-functional project via bounding box error
If you click on a project under Select Project, then draw a new bounding box, then go back and select None under Select Project, the old bounding box remains and you can draw a second (see image). If you then create the project, it has no map. When you publish the project, it creates an error only, there is no project (see image). See project here: https://collect.earth/collection?projectId=6378

biplovbhandari:
closes #435


sirmspencer:
This error is because the base map is incorrectly set to blank.  Please fix this instead of waiting for the incorrect values appear.  Go to setProjectTemplate starting on line 242 and set correct values for when newTemplateId === 0.  You should also clear the bounding box on the map if the coordinates are clear.

PR #440, opened 2020-04-02
arrange the planet daily layer in the descending order in the toggle on/off layer list

biplovbhandari:


lambdatronic:
mercator-openlayers.js:
- Lines 160-173: Rewrite this in functional style:
```js
mercator.getTopVisiblePlanetLayerDate = (mapConfig, layerTitle) => {
    const layer = mercator.getLayerByTitle(mapConfig, layerTitle);
    if (layer && layer instanceof LayerGroup) {
        const planetLayers = [...layer.getLayers().getArray()];
        const visibleLayer = planetLayers.reverse().find(planetLayer => planetLayer.getVisible());
        return visibleLayer ? visiblelayer.get("title") : "NA";
    } else {
        return "NA";
    }
};
```
- Line 235: Don't mutate. Rebind:
```js
const sortedData = data.filter ...
```
- Line 237: Use ES6 arrow syntax for anonymous functions.
- Lines 244,247: data -> sortedData
- Lines 263: layerGroup.getLayers().getArray().reverse()

sirmspencer:
src/main/resources/public/js/mercator-openlayers.js
Line 240 remove console log
Line 259, You should not need to spread the array back into an array if you are not adding and element.

line 407 this.setState({ loading: this.state.currentImagery.sourceConfig.type === "PlanetDaily" });

biplovbhandari:
mercator-openlayers.js Line 259 While it's true that I don't need spread operator in case not modifying element, I remember last time when I tried this, it would display the layers in ascending order as I passed to the PlanetLayerSwitcher using layerGroup.getLayers().getArray().reverse(). Now it looks like the access token has expired and I am getting nothing back from gee-gateway in order for me to test again.

sirmspencer:
Im not sure the spread is still appropriate here.  Reverse is a mutating function, so you created a shallow copy before reversing.  The problem is that in PlanetLayerSwitcher you reference the layers (`layer.setVisible`, etc).  Im not sure that the reference you are passing is correct anymore.

biplovbhandari:
@sirmspencer done

PR #443, opened 2020-04-03
Fixes window size & menu failure

### LINKED ISSUE #421: window size & menu failure
Reducing window size below 958 pixels kicks in adaptive site design where only the Institution list is shown. (this is also the default state for mobile)

When the adaptive site design is active, there are host of issues:
- three horizontal line menu icon does not work on any page
- right hand Plot Navigation etc. menu does not appear when in a project
- when on the index page, the map does not appear (this may be intentional)

rfontanarosa:
Fixes #421 

fixed collapsed navigation bar (adding bootstrap)
fixed unreachable sidebar (adding a button that allows to switch between map and sidebar)


sirmspencer:
Line 40, this is a boolean state.  Use true / false :showSidebar instead of text
Line 1157 showMapOrSidebar -> toggleShowSidebar
Line 1158 this.setState({ :showSidebar !this.state.showSidebar });
Line 1161, you should not need this in react
Line 1172 This line is too long.  Move onClick, className and style onto their own line. Move the individual style tag to their own line.
Line 1172 use the classes `btn bg-lightgreen` instead of `height: "3rem", width: "3rem", backgroundColor: "#31BAB0", borderRadius: ".2rem"`.  Make right 2rem, the analysis pane has its own scroll bar that interferes with the alignment
Line 1773, you may need to adjust the margin after the above changes.
Line 1174 `{this.state.showSidebar ? <UnicodeIcon icon="downCaret"/> : <UnicodeIcon icon="upCaret"/>}`

src/main/resources/public/js/jquery-3.1.1.min.js
remove the old file

src/main/resources/public/js/bootstrap.min.js.map
src/main/resources/public/js/jquery-3.1.1.min.map
I dont think we need the map files

src/main/resources/public/js/bootstrap.min.js
What is this needed for?  Everythings to work for me without it.  It doesnt look like its being used in the code.

src/main/resources/template/freemarker/header.ftl
Line 31, remove uneeded dependency

sirmspencer:
Can you be more specific as to what exactly does not work without bootstrap.js?  I can not see any difference.

PR #448, opened 2020-04-07
Fixes weird plot distribution & no sample points in plots 

### LINKED ISSUE #129: Allowing for one sample per plot in the center
I am trying to classify Landsat pixels and assign just one value (30 meter plots with one sample); however, the samples seem to start from a plot corner. Can we please add functionality to add just one sample per plot in the center on the project design page?

### LINKED ISSUE #439: Weird plot distribution & no sample points in plots
https://collect.earth/collection?projectId=6483

While testing the behavior of the sampling plot/points for the manual, I ran into some strange behavior in this project:
There are no sample plots on the right hand side of the sample area. 
There are no sample points in each sample plot.

biplovbhandari:
Fixes #439 fixes #129 

no sample points in plots


biplovbhandari:
@lambdatronic 

Steps to reproduce sample plots gridded 500 m apart, 40m circles. Sample points gridded 30m.

This is a very edge case. If we comment [this line](https://github.com/openforis/collect-earth-online/blob/master/src/main/java/org/openforis/ceo/utils/ProjectUtils.java#L386), the error is clear with the produced samples as shown in below figure. Meaning the sample points lie outside of the plot area. So decreasing the padding distance to include at least one of the sample points.

image

sirmspencer:
Your fix just shifts the points around and you get one left in the corner.  This is not particularly desirable either.  Lets do the math on the max gridded spacing you can have for a circle and prevent the user from making bad plots.  We do something similar to prevent a user from created too many plots or samples.

sirmspencer:
For a square plot the the sample resolution must be < width.  For a round plot, the sample resolution must be less than diameter / square root 2.  For both plots, allow the resolution to be > twice the width / diameter (for now it is a hack to get one sample in the middle.

biplovbhandari:
@sirmspencer please review

sirmspencer:
I like this solution much better, it is going to be more intuitive to the user.

For center, set the sampleResolution to 2 * the plot value when the form is submitted.  This is for two resons, 1. right now if the user changes from random to gridded, we just disable the parameter that is no longer used, that means if you switch back your value remains.  Calculating the sampleDistribution when switching to center changes the users previous input. 2. it will take way less lines of code for that solution.
130, `sampleDistribution: this.state.projectDetails.sampleDistribution === "center" ? "gridded" : this.state.projectDetails.sampleDistribution,`
132, `sampleResolution:  this.state.projectDetails.sampleDistribution === "center" ? 2 * this.state.projectDetails.plotSize :  this.state.projectDetails.sampleResolution,`
318-319, 433, 485, 486, 825, this is not needed as a separate function.  
603, 606, 809-813, 828, remove
882, only set the sampleDistribution

PR #450, opened 2020-04-10
Add option in project setup to add button to load the Collect Earth desktop GEE script

### LINKED ISSUE #449: add option in project setup that adds a button to load the Collect Earth desktop GEE script
Add an option in project setup that an administrator can choose to add a button in the data collection page that opens the GEE script that is used by the Collect Earth desktop version. Billy and John looked at optimizing the script so it loads faster, here's the script: https://billyz313.users.earthengine.app/view/ceoplotancillary#geoJson=%7B%22type%22:%22MultiLineString%22,%22coordinates%22:%5B%5B%5B100.0,%200.0%5D,%5B101.0,%201.0%5D%5D,%5B%5B102.0,2.0%5D,%5B103.0,3.0%5D%5D%5D%7D

Finally, Billy mentioned the next step was to add a button to the collection page that opens the link and passes the correct geometry in the geoJson parameter.

(I am hoping this is not a duplicate issue request, but I didn't see it in list)

biplovbhandari:
Fixes #449


sirmspencer:
- Please make these same changes in src/main/java/org/openforis/ceo/local/JsonProjects.java

- Line 43/171, this is not needed, projectOptions are part of the currentProject state object
- Line 1294, `projectOptions={this.state.currentProject.projectOptions}`
- Line 1565, window.open needs a window name so the window can be reused each time the user clicks the button.
- Line 1603, this line is too long, put returns before each &&

- Line 39/40, Do not maintain this data twice.  Either maintain the data in the projectOptions state object or create a new object from the individual state variables before submitting the api.
- Line 407, this is not valid, this.state.showGEEScript is never changed.  You are modify only the projectOptions. If you use the projectOptions object you would need to do `this.setState({ projectOptions: {...this.state.projectOptions, showGEEScript: !this.state.projectOptions.showGEEScript }})`

src/main/resources/sql/update_2020-04-10.sql
- Please make these same changes in src/main/resources/sql/load_ceo_functions.sql

biplovbhandari:
Collection.js Line 1561 adding a name to window.open does not seem to have an effect on my PC. It changes the URL but doesn't send a new request. Not sure why this is the case

sirmspencer:
> Collection.js Line 1561 adding a name to window.open does not seem to have an effect on my PC. It changes the URL but doesn't send a new request. Not sure why this is the case

window.open is working correctly after adding a title.  the URL is not changing because the geojson you are sending is for the entire project area and not the plot.

biplovbhandari:
@sirmspencer  I think it does gets the extent of the current map view and not the entire project area from [here](https://github.com/openforis/collect-earth-online/blob/master/src/main/resources/public/js/mercator-openlayers.js#L84). The URL does changes but the map view does not change, i.e. even the URL(geojson) is changed there is no reloading taking place on the external URL.

sirmspencer:
I see, I miss interpreted because Earth engine is showing a rectangle the shape of the project instead of a circle.  However the size and extent look correct.

image

I see now, the URL is is changing but earth engine doesnt want to refresh.  Look at the parameter `replace` https://www.w3schools.com/jsref/met_win_open.asp

sirmspencer:
Ok, this window thing was a little tricky because its cross origin.  This should work
```
constructor(props) {
        super(props);
        this.state = {
            auxWindow: null,
            newPlotInput: "",
            showNav: true,
        };
    }
```
and
```
    loadGEEScript = () => {
        const geometry = mercator.getViewPolygon(this.props.mapConfig);
        const geoJson = "{\"type\": \"Polygon\", \"coordinates\":" + JSON.stringify(geometry.getCoordinates()) + "}";
        this.state.auxWindow && this.state.auxWindow.close();
        this.setState({ auxWindow: window.open("https://billyz313.users.earthengine.app/view/ceoplotancillary#geoJson=" + geoJson, "_ceo-plot-ancillary") });
    };
```

biplovbhandari:
@sirmspencer please review

PR #451, opened 2020-04-14
Fix miscounted analysis column in download plot

### LINKED ISSUE #445: Miscounted ANALYSES column in Download Plot Data.
The ANALYSES column is being miscounted in Download Plot Data. When the backend algorithm summarizes the points for each plot, it looks like it's counting 0s as 1 (like, as TRUE there is something in the field) or something.

When no one has analyzed a plot, the field equals the number of samples per plot.

image

When someone has analyzed the plot, it still equals the number of samples per plot.

image

biplovbhandari:
Fixes #445

fix miscounted analysis column in download plot


sirmspencer:
Does this error affect the JSON database?

- Line 71, this change does not belong in this PR

src/main/resources/sql/update_2020-04-14.sql
- Please make the same changes in src/main/resources/sql/load_ceo_functions.sql

biplovbhandari:
I don't think it affects the JSON DB, there is no analysis column.

sirmspencer:
1861, you changed the order of the columns in the select. They no longer match the order of the function definition return.
This change will not work for older projects, we only started collecting the time recently. Use `user_rid` instead of `collection_time`.

biplovbhandari:
@sirmspencer There is no `user_rid`, instead there is `username` in the join. Please review

sirmspencer:
Your right, I was looking at the user_plots table not select_all_project_plots.  Its been a while since ive looked at these so my memory isnt perfect.  I took another look and select_all_project_plots already returns the correct value for flagged and assigned for each plot.  The problem is then we inner join with samples so that data repeats X times. 

So it looks like flagged and assigned both just need to be:
```
MAX(flagged) as flagged,
MAX(assigned) as assigned,
```

PR #454, opened 2020-04-22
a single filter for both affiliations and other institutions

biplovbhandari:
a single filter for both affiliations and other institutions

### DIFF
--- a/src/main/resources/public/jsx/home.js
+++ b/src/main/resources/public/jsx/home.js
@@ -211,24 +211,17 @@ class SideBar extends React.Component {
         super(props);
         this.state = {
             filterText: "",
-            filterTextUserInstitution: "",
             filterInstitution: true,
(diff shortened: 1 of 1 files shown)


sirmspencer:
Please run eslint on this code.

sirmspencer:
Lines 237-239, remove this header to save space
Line 241, move the create button under the institution filter
Line 343, continue this styling so that the affiliation selector is not just the size of one institution.

biplovbhandari:
Regarding Line 237-241, I think we still need it for public users and super admin (as admin list is not shown in filtered). Instead, I have added the check to show only appropriate to save the space

I don't quite get the Line 345 styling one as this is applied to both affiliation and other institutions from Line 245 and 265.

sirmspencer:
"your affiliations" only show 1 row.  We need to make this section bigger.
image

PR #456, opened 2020-04-24
Enable dashboard configuration copy from another project

### LINKED ISSUE #280: GEODASH - Allow users to copy widgets from another project
Create a new route, connect it to the dashboard edit page

billyz313:
This closes #280 and allows users to copy the dashboard from any other project they have access to.

### DIFF
--- a/src/main/resources/public/jsx/widget-layout-editor.js
+++ b/src/main/resources/public/jsx/widget-layout-editor.js
@@ -14,7 +14,7 @@ class BasicLayout extends React.PureComponent {
             selectedProjectId: 0,
             projectList: [],
             projectFilter:"",
-            isEditing: false,
+            addDialog: false,
             copyDialog: false,
             addCustomImagery: false,
             selectedWidgetType: "-1",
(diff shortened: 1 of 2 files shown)


sirmspencer:
- Line 17 isEditing -> addDialog (global replace)
- Line 54 remove extra space
- Lines 894-917 This is the same functionality you have in componentDidMount.  Pull this out into a function for reuse.  A function can return a promise.  You would then use it like this.getWidgets().then.catch or this.getWidgets().catch depending on what you need. 
- Lines 918-920, remove extra spaces around parens
- Line 945, dont use alert here, this is being called inside a loop and a bad connection could lead repeated diolague boxes. Also you message is coppied from a different funtcion.
- Line 949, 997, 1309, 1335, 1343 parans around a contitional statement are not needed
- Line 988 - 990, make this one line like you do for the other function components.
- Line 1022 widgetTypeSelect -> project-template
- Line 1044, It was unclear to me what this meant just reading it.  Consider rephrasing to something like "Warning, selecting a template project will overwrite existing widgets immediately."
- Line 1091-1095, 1113-1117, 1352-1354 This should be part of the onWidgetTypeSelectChanged event for the widget type and not part of the render function.
- Line 1110, after above this is the same return value as 1088
- Line 1918 "Copy Layout"

src/main/resources/template/freemarker/geo-dash-navbar.ftl
Lines 52/53 Remove extra space before )

sirmspencer:
- Line 54, 906 remove extra space
- Line 308, `imageCollection: event.target.value === "ImageElevation" ? "USGS/SRTMGL1_003" : ""`
- Line 330 The result of the conditional is already a boolean. 
	`const formReady = event.target.value === "statistics"
			            || event.target.value === "imageAsset"
			            || event.target.value === "imageCollectionAsset"
			            || event.target.value === "ImageElevation"`

 or more simply
	`formReady: event.target.value === "statistics"
	            || event.target.value === "imageAsset"
	            || event.target.value === "imageCollectionAsset"
	            || event.target.value === "ImageElevation"`
- Line 292-302, remove after adding above. 
- Line 395, 647, 939, 1862 isEditing -> addDialog

PR #465, opened 2020-05-17
allow review of project options

### LINKED ISSUE #462: Show project options on the review project page
In #450 the option was added for new projects.  Show this option on the review projects page.  Allow the user to change this option.

biplovbhandari:
fixes #462


sirmspencer:
For update_project, follow the same pattern you used in create project, and pass the project_options json.

biplovbhandari:
@sirmspencer please review

sirmspencer:
Line 258, event is not used
Line 264, 265, 266 missing trailing comma (eslint rule 'comma-dangle')

biplovbhandari:
@sirmspencer Please review

sirmspencer:
src/main/java/org/openforis/ceo/local/JsonProjects.java
Lines 745-747, 1320-1322

515-514, 909-911

You do not need to create a json object just to convert it back to a string.  Use getOrEmptyString().getAsString() == "" for checking for valid data because it checks for more than just an empty value. We could update that function to take an optional fall back string, but lets not do that in this PR.  Your final versions for all four locations should look more like this:
```
final var projectOptions = getOrEmptyString(jsonInputs, "projectOptions").getAsString();
pstmt.setString(6, projectOptions == "" ? "{\"showGEEScript\":false}" : projectOptions);
```

PR #466, opened 2020-05-18
change project table to include imagery rid

biplovbhandari:
load and update imagery rid on the project creation/update page


sirmspencer:
You no longer need the text field base_map_source in the projects table.  Drop all references to it in the table, create_project, update_project, etc.  Create a sql update to convert the current base_map_source into the imagery_rid field.

biplovbhandari:
@sirmspencer Please review

sirmspencer:
Line 936, you cant send 0 to PG, imageryId has a foriegn key and there is no 0. You will need to create a function to find the first publicly avaiable imagery and use that id.  When you are done verify that if no imageryId is sent that there are no errors.

Line 14, imageryId is a numeric, use -1
Line 169, imageryId is a numeric, use `{imageryId || -1}`
Line 242, imageryId is a numeric, check for a positive number. 
Line 73/75 update mercator to use id instead of title.

Line 55, put `this.state.plotList` on its own line
Lines 44/45, 226/227 update mercator to use id instead of title.

src/main/resources/sql/update_2020-05-18.sql
Lines 6-9, this is not complete.  There can be imagery with the same name if the institution is different. Do you have a copy of the collect.earth db to test this on locally?  There may be cases where bad data exists, since there is no unique constraint actually set. Your update could fail if your from / where clause produces more than one result.
Line 120, do you need to drop the view before recreating it?

biplovbhandari:
@sirmspencer I do not have a copy of the db. Would it be possible for you to send me a copy over to my email? Thanks~

biplovbhandari:
@sirmspencer Please review. 

1. I saw that there are some rows of `base_map_source` as `PlanetGlobalMosaic` which are not present in the imagery as `public` nor as to the `private` source for the institutions. This might indicate that at some point those were `public` and then turned to `private`. So these records were not converted. Others are doing fine as per the DB that you shared. I might need your help with updating those. But please test on your local repo before trying out in the production.

2. I think updating Mercator to use id instead of title can be a separate issue that I can work on after this is merged.

Thanks!

sirmspencer:
There is an imagery for PlanetGlobalMosaic for institution_rid = 3.  institution_rid = 3 is SIG, where most public imagery is hosted.  It is not private because we retired it.  If any projects use PlanetGlobalMosaic but their institution does not match any other available imagery use the one that is institution_rid = 3.

sirmspencer:
I will help you with the queries to update the imagery.

biplovbhandari:
@sirmspencer 
1. I just updated the PR to include `PlanetGlobalMosaic` from `institution_rid` 3. But there is still one record from the backup DB that has `PlanetDaily` `base_map_source`, which is not included in the `institution_rid` 3, rather on another `institution_rid` 226. Please make the adjustment to which you think is best.
2. I think that `DROP COLUMN base_map_source CASCADE` drops the view as well. So haven't included it.

Again please test on your local repo before trying out in the production. Thanks!

sirmspencer:
509-513, use a complete if else statement instead of mutating a variable.
```
if (rs.next()) {
    return rs.getInt("select_first_public_imagery");
} else {
	return 0;
}
```

63, Dont drop the base_map_source just yet.
58, move imagery_id to the end so that it matches our live database.  ALTER TABLE will add the new column to the end.

src/main/resources/sql/update_2020-05-18.sql
32-66, This simplifies. After you have set all of the valid base_map_sources, just set all remaining null imagery_rid to select_first_public_imagery
56-66, Sorry I misspoke a little bit.  While its true that there is still a reference to PlanetGlobalMosaic in institution 3, it is not a valid choice moving forward.  It can just be set to select_first_public_imagery
Dont drop the base_map_source just yet.  Lets give it some time, just in case the update is not sucessful

PR #472, opened 2020-05-28
SAR integrated to degradation widget

billyz313:
SAR integrated to degradation widget and added window name to collection page to eliminate navigation issue.

### DIFF
--- a/src/main/resources/public/css/geo-dash.css
+++ b/src/main/resources/public/css/geo-dash.css
@@ -301,6 +301,7 @@ input:checked+.switchslider:before {
     color: white;
     line-height: 26px;
     cursor: default;
+    min-height: 26px;
 }
 input#widgetDualLayer {
     display: inline-block;
--- a/src/main/resources/public/jsx/collection.js
+++ b/src/main/resources/public/jsx/collection.js
@@ -52,6 +52,7 @@ class Collection extends React.Component {
     }
 
     componentDidMount() {
+        window.name = "_ceocollection";
         this.getProjectData();
     }
 
--- a/src/main/resources/public/jsx/geo-dash.js
+++ b/src/main/resources/public/jsx/geo-dash.js
@@ -386,14 +386,12 @@ class DegradationWidget extends React.Component {
         this.state = {
             selectedDate: "",
             degDataType: "landsat",
-            graphLoading: true,
         };
     }
 
     handleDegDataType = (dataType) => {
         this.setState({
             degDataType: dataType,
-            graphLoading: true,
         });
     };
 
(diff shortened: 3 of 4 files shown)


sirmspencer:
DegredationWidget, it looks like you are trying to set graphLoading to true in GraphWidget widget when degDataType.  DegradationWidget does not need its own graphLoading state.  Create one function to update the degDataType state in DegradationWidget.  In GraphWidget watch for degDataType in component did update and set the graphLoading state in GraphWidget accordingly. 

MapWidget
componentDidUpdate provides prevProps and prevState, you dont need to keep separate state to see if something is changing. Look at collection.js for an example. It looks like wasFull, lastDegDataType, lastStretch, and selectedDate can be removed from the state.
Line 477, 478, 480, 481, remove
Line 680 `componentDidUpdate(prevProps, prevState)` 
Line 681 `this.props.widget.isFull !== this.prevProps.widget.isFull`
Line 683 remove
Line 684, line after if statements if they are not related.
Line 685 should this be `this.props.mapCenter !== this.prevProps.mapCenter`
Line 688 `this.props.selectedDate !== this.prevProps.selectedDate`
Line 689 `this.state.stretch !== this.prevState.stretch`
Line 690 `this.props.degDataType !== this.prevProps.degDataType`
Line 710 remove


Lines 1181-1199 
```
    toggleStretch = evt => this.setState({ stretch: evt.target.checked ? 543 : 321 });

    toggleDegDataType = evt => this.props.handleDegDataType(evt.target.checked ? "sar" : "landsat");
 ```

Lines 1191-1199, you dont need a separate function, just handle this correctly in DegradationWidget.

GraphWidget
Line 1254 componentDidUpdate provides prevProps and prevState, you dont need to keep separate state to see if something is changing. 
Line 1257 just set the local graphLoading state.  There is no need to have the DegredationWidget pass a prop.
Line 1271, 1271 make all the : lines have the same indent.  It is easier to see that there is not multi level nesting of the conditional and is just a chain of if->then like a switch statement.
Line 1302, theKeys is only used in the else block.  Move the definition to the same line as assignment, 1321 (the way you had it before.)
Lines 1367/1368 You do not need to store values in state to find changes in componentDidUpdate
Lines 1371-1374 All you need is `this.setState({ graphLoading: false }`

**Comment by sirmspencer** [comment]
Line 389, 396, graphLoading is not used.
Line 444 handleDegDataType is not used in the GraphWidget
Line 464 lastStretch is no used
Line 1216, this line is too long, use
```
if (this.props.degDataType
    && (prevProps.degDataType !== this.props.degDataType
        || prevState.selectSarGraphBand !== this.state.selectSarGraphBand)) {
```
Lines 1297, 1298, Have || start the line not end it.
Line 1436, you dont need `=== true`, it is already boolean

PR #474, opened 2020-05-29
Fix css on popup

sirmspencer:


lambdatronic:
imagery-list.json:
- Update title and attribution fields for Mapbox layers

PR #483, opened 2020-06-08
mercator to use id instead of title

biplovbhandari:


sirmspencer:
src/main/resources/public/js/mercator-openlayers.js
 647, 662, 672, 673, 684, 721, 875, 1051 Update comments

PR #508, opened 2020-06-10
wri securewatch request changes

biplovbhandari:


sirmspencer:
review-institution
Line 666, can this be generic?
```
} else if (sourceConfig.startDate
            && sourceConfig.endDate
            && (new Date(sourceConfig.startDate) > new Date(sourceConfig.endDate))) {
    return "Start date must be smaller than the end date.";
} else {
```

biplovbhandari:
@sirmspencer 

1. If there is no date in the range, the mosaic image would be shown and the date filter would not be available.

2. src/main/resources/sql/update_2020-06-10.sql: It does append in the existing source_config.

3. Yes I think it can be generic. Thanks! I have updated the PR.

PR #510, opened 2020-06-10
Webpack clean up

sirmspencer:
Single config file, pass env
Filter .map files


lambdatronic:
webpack.config.js:
- Lines 10,133: Do we need the enclosing () here since this function simply returns a value?

sirmspencer:
> webpack.config.js:
> 
> * Lines 10,133: Do we need the enclosing () here since this function simply returns a value?

Yes, to return an object it needs to be wrapped in ().  Otherwise it tries to interpret the { as the start of a function body.

PR #511, opened 2020-06-11
link imageries to projects and use them in the collection page

biplovbhandari:


sirmspencer:
- 170, 191, Put these behind the same conditional (imageryList), and show "Loading Imagery..." while the imagery is loading.
- 197, do not pad { with a space in JSX blocks.
- 201 Include imagery.id in your closure instead of discovering it later. onChange={e => addRemoveProjectImagery(imagery.id, e.target.checked)}
- 203 Show the imagery as checked if it is selected as the default.  Also disable the input so the user cant unselect it.

- 422 - 453 this is repeat code.  Move the helper functions locally to where they are needed in `<ProjectAOI>`.  Make the function here a simple setProjectImagery(newProjectImageryList)
- 450 When you move it to `<ProjectAOI>` use
```
addRemoveProjectImagery = (imageryId, add) =>
    if (checked) {
    	this.setState({ projectImageryList: [...this.state.projectImageryList, imageryId]
    } else {
        this.setState({
            projectImageryList: this.state.projectImageryList.filter(imagery => imagery !== imageryId),
        })
    });
 ```
 - 477, 519 setProjectImagery={this.setProjectImagery}
 - 511 Do not filter the imagery list at this time.  The issue with what you have, is that if the user changes between public and institution, the selected imagery is not cleared or updated. Lets move this functionalit to a new PR and discuss in mattermost with Gary.

- 294-307 Same as with create-project: this is repeat code.  Move the helper functions locally to where they are needed in `<ProjectAOI>`.
- 321, 516 setProjectImagery={this.setProjectImagery}

- 606 add return back before UNION
- 627 Why do you need DISTINCT here?  There are no duplicates in imageries, and there should be no duplicates in project_imagery.  A proper join should result in no duplicates.
- You will need to update the sql function archive_imagery to remove archived imagery from project_imagery
- select_imagery_by_project We need to discuss in mattermost with Gary whether we should always be returning public imagery or only the selected imagery.  If we are only returning the project_imagery, we will need to decide what happens if a user does not have the rights to any of the project_imagery.  This I think will be the same conversation as my comment about not filtering the imagery in create-project.js. It should be fine for now, we will address this in a new PR.

**Comment by sirmspencer** [comment]
- 453 - 464 this is repeat code.  Move the addRemoveProjectImagery to `<ProjectAOI>`.
- 473, 510 setProjectImagery={this.setProjectImagery}

- 204-300 same as with create-project: this is repeat code.  Move the addRemoveProjectImagery to `<ProjectAOI>`.
- 315, 510 setProjectImagery={this.setProjectImagery}

Both create and update project need to be protected against a call with no "projectImageryList"
- 560-562 move this code into insertProjectImagery(), 
- 562 you dont need to add imageryId with the changes you made to SQL. This will help, you dont need to pass imageryId to insertProjectImagery
- 547-550 I think you can revert this after the comment above
- 563 insertProjectImagery(projectId, jsonInputs);
- 937 I think you can revert this because we dont need to create a new array with imageryId
- 939 this whole section is a bit of a relic.  Some of the parameters are used later which is why the object exists.  projectImageryList is not used later so you dont need to add it to the newProject object.
- 1014 insertProjectImagery(newProject.get("id").getAsInt(), jsonInputs);

- 106 setProjectImagery
- 173 / 200, combine these and only show one loading message.
- 209-223 You can do this with one input.  Disabled already prevents the event from happening.
```
<input
    className="form-check-input"
    id={imagery.id}
    onChange={e => addRemoveProjectImagery(imagery.id, e.target.checked)}
    type="checkbox"
    disabled={imagery.id === imageryId}
    checked={projectImageryList.includes(imagery.id) || imagery.id === imageryId}
/>
```

- 92, create a unique constraint on imagery_rid and project_rid
- 635 `AND pi.project_rid = p.project_uid`
- After the above, you should be able to SELECT without distinct.

PR #516, opened 2020-06-11
Split imagery routes for better authentication

sirmspencer:


lambdatronic:
Server.java:
- Unused: /get-institution-imagery

JsonImagery.java:
- Lines 20,26,32: Remove @Override Java metadata tags.
- Line 38: Drop getAllImagery.

PostgresImagery.java:
- Line 29: Why aren't we removing access tokens from source_config
  anymore? This seems like a security hole that negates the point of
  imagery proxying.

update_2020-06-09.sql:
- There is already a file with this name on master. Rename this file
  to a later date.

These files all call /get-all-imagery and must be updated in this PR:
- create-project.js
- geo-dash.js
- project-dashboard.js
- review-institution.js
- review-project.js
- widget-layout-editor.js

lambdatronic:
PostgresImagery.java:
- Line 27: There is no longer a featureProfile attribute, so this line can be dropped.
- Line 76: This looks backwards. You should be setting stripConfig = true is isInstAdminQuery() returns false. So use `not` here.

PR #518, opened 2020-06-15
Update imagery

### LINKED ISSUE #398: Allow users to edit their institution's imagery sources (and make sure they appear on the Review Project page)
There is no way to edit the imagery source. The process of adding WMS services is very error prone (e.g. mistyping the JSON parameters or changing the resolution or format) so you should be able to edit the imagery sources (maybe the base ones like Bing Maps or Digital Globe should be " locked" )
http://i.imgur.com/0jW6gaw.png

When I choose the PNOA as a basemap it works fine assessing plots, but if I go back to the project page (using the Edit button) the basemap parameter appears empty, even though PNOA is in the list of avaliable sources. see images.

http://i.imgur.com/zurwJEF.png
http://i.imgur.com/gPUGd6u.png

rfontanarosa:
Closes #398


sirmspencer:
A couple general comments:
- We are trying to make this site as small as possible for people with low speed internet connections.  Please avoid importing new npm packages unless absolutely necessary.  I dont think Accordion is needed here, the hide show functionality is simple and there are plenty of other examples in the code base.
- The changes you made to the layout make sense to me.  Can you please take some screen shots to get general approval?
- I spend a good amount of time making the new imagery component generic so any additions to imageryOptions would instantly work in the UI.  The same principles need to be applied when editing the imagery.

sirmspencer:
- 86, 106 In general dont use a different name for props than the variable you are passing so `showImagery={this.state.showImagery}` `showUsers={this.state.showUsers}`
- 326 imageryToEdit is set to an opject that is the imagery.  The default value needs to be something like `{id: -1}` (or imager_id, what ever matches )
- 407 Move titleIsTaken back to the ImageryList component and pass it instead of this.state.imageryList. imageryList is not used outside of titleIsTaken in NewImagery
- 685 - 706 This is repeat code with addCustomImagery, refactor into a validateImageryParams() function

src/main/resources/public/css/custom.css
- 560 remove accordion__button

sirmspencer:
This is superseded by #549

PR #519, opened 2020-06-15
Allow more csv columns2

### LINKED ISSUE #408: Upload csv: allow more columns than just lon, lat, plotid
Save columns in the export as well.

rfontanarosa:
Closes #408

### DIFF
--- a/src/main/java/org/openforis/ceo/local/JsonPlots.java
+++ b/src/main/java/org/openforis/ceo/local/JsonPlots.java
@@ -95,6 +95,8 @@ public class JsonPlots implements Plots {
                                     .findFirst();
 
         if (matchingPlot.isPresent()) {
+            // TODO add an object which contains additional information
+            matchingPlot.get().add("extraPlotInfo", new JsonObject());
             unlockLockPlots(projectId, matchingPlot.get().get("id").getAsInt(), userId);
             return matchingPlot.get().toString();
         } else {
--- a/src/main/java/org/openforis/ceo/postgres/PostgresPlots.java
+++ b/src/main/java/org/openforis/ceo/postgres/PostgresPlots.java
@@ -38,7 +38,12 @@ public class PostgresPlots implements Plots {
     private static JsonObject buildPlotCollectionJson(ResultSet rs) {
         var singlePlot = buildPlotJson(rs);
         try {
-            singlePlot.addProperty("extraPlotInfo", rs.getString("extra_plot_info"));
+            var extraPlotInfo = parseJson(rs.getString("extra_plot_info")).getAsJsonObject();
+            extraPlotInfo.remove("gid");
+            extraPlotInfo.remove("lat");
+            extraPlotInfo.remove("lon");
+            extraPlotInfo.remove("plotid");
+            singlePlot.addProperty("extraPlotInfo", extraPlotInfo.toString());
         } catch (SQLException e) {
             System.out.println(e.getMessage());
         }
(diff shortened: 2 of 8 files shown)


sirmspencer:
We are not displaying extra fields for samples, and I dont think we should.  Please remove the joins with the external sample tables and do not return samples.rem_data.

rfontanarosa:
> You are not displaying extra fields for samples, and I dont think we should. Please remove the joins with the external sample tables that you added and do not return samples.rem_data.

Ok!

rfontanarosa:
Hi @sirmspencer,

Now I remember why I used org.apache.commons.csv library.

JsonProjects doesn't support extra fields and the CSV parsing (line.split(",")) has been written to extract only the plot center and it's not compatible with escaped fields, that's why I needed a complex library.

Anyway, this is what I wrote: https://github.com/openforis/collect-earth-online/commit/451958bf.

sirmspencer:
> Hi @sirmspencer,
> 
> Now I remember why I used org.apache.commons.csv library.
> 
> JsonProjects doesn't support extra fields and the CSV parsing (line.split(",")) has been written to extract only the plot center and it's not compatible with escaped fields, that's why I needed a complex library.
> 
> Anyway, this is what I wrote: [451958b](https://github.com/openforis/collect-earth-online/commit/451958bf).

sirmspencer:
src/main/java/org/openforis/ceo/postgres/PostgresPlots.java
- 38-47 Go ahead and make a copy of `buildPlotJson` here and then remove the columns that are unused by plot collection routes.
- 41 extraPlot -> extraPlotInfo, extra_plot -> extra_plot_info
- 41 remove th
- 83 / 125, you can move adding samples into buildPlotCollectionJson

- 1335-1341 this can fit on less lines. Pass only the needed props 
```
{this.state.currentPlot
    ? <PlotInformation extraPlotInfo={this.state.currentPlot.extraPlotInfo}/>
    : null
}
```
- 1670-1676 move this function to utils/textUtils.js
- 1676 extraPlot -> extraPlotInfo
- 1679, 1680 You do not need to deref here, since it does not simplify the code.
- 1681 Instead of this, make sure the back end always returns and object, even if it is empty.
- 1686 showGroup={this.state.showInfo}
- 1687 you only need to reference previous state if you are making multiple state changes at once.  Just use `toggleShow={() => this.setState({ showInfo: !this.state.showInfo })}`
- 1691, 1693 parsedExtraFields -> this.props.extraPlotInfo
- 1693 JSK brackets are not padded with spaces. You also have an extra space before your last `)`
- 1696 "There is no extra information for this plot."

- 339 extra_plot -> extra_plot_info

src/main/resources/sql/update_2020-06-10.sql
- Rename this file, there is already a file with that name.  We really should start adding the a comment to the end of the name like `update_2020-06-10_csv_cols.sql`
- 24 extra_plot_info

rfontanarosa:
buildPlotJson is used by getProjectPlots.

getProjectPlots is used by "get-project-plots" endpoint.

"get-project-plots" endpoint is used by:

- collection.js
- create-project.js
- project-dashboard.js
- review-project.js

the plot list returned is used as argument to call these mercator functions:

- addPlotOverviewLayers
- plotsToVectorSource
- addPlotLayer

the attributes used inside these functions are:

- flagged
- analyses
- id
- center

I assume I have to reduce the attribute list to the four above. :)

sirmspencer:
Data from a shp file can contain objects.  Filter these and nulls. collections 1682
```
{Object.entries(parsedExtraFields)
                                .filter(([key, value]) => value && !(value instanceof Object))
                                .map(([key, value]) => <li key={key}>{key} - {value}</li>)}
```

PR #521, opened 2020-06-16
Store chart data, add band options, fix sample image layout

billyz313:
Store chart data after initial load of each data type, add band options for Landsat visualization, fix sample image layout to match actual layout.


sirmspencer:
- 1250, this line is way to long
```
graphRef: this.createChart(
    widget.id,
    widget.graphBand != null ? widget.graphBand : indexName,
    this.state.chartData[this.props.degDataType].data,
    widget.graphBand != null ? widget.graphBand : indexName
),
```
- 1250, is ok for widget.graphBand to be undefined? If not use
```
graphRef: this.createChart(
    widget.id,
    widget.graphBand || indexName,
    this.state.chartData[this.props.degDataType].data,
    widget.graphBand || indexName
),
```
- 1250, 1354 why are you passing the same value to createChart in two parameters?
- 1232, 1461 Length cant be less than 0. What are you checking for here?
- 1321 I think you are missing some parentheses.  this.props.degDataType !== "sar" || this.props.degDataType === "sar" is always true.

---------  A group of related changes
- 1286 this is not scoped where it is used
- 1290 - 1294, this can be done simpler with filter and map.
- 1295 Remove. Dont assign a value to itself.  Also sort() mutates the original array so it doesnt need to be reasinged.
- 1298 how can this have a value if Object.keys(res.timeseries[0][1]).length === 0
- 1299 data: res.timeseries.filter(v => v[0]).map(v => [v[0], v[1]]).sort((a, b) => a[0] - b[0]);
- 1367 Remove unneeded function.
----------

billyz313:
1298 Object.keys(res.timeseries[0][1]).length === 0 is checking to see which format is being returned.  This is a format that is returned with no keys.

sirmspencer:
src/main/resources/public/locale/geodashhelp_en.json
src/main/resources/public/locale/geodashhelp_es.json
40-42, remove space after : for consistency

PR #523, opened 2020-06-17
fix quit button on the collection page

biplovbhandari:


sirmspencer:
1284-1288, you only need prevState if you are manipulating the same state multiple times.  Use `toggleQuitModal = () => this.setState({ showQuitModal: !prevState.showQuitModal });`
2062-2071 This is overly complicated.  You should be able to acomplish the same thing by setting the content div with a higher z-index.  The on-lick of the div with a lower z-index will not fire for click on a div with a higher z-index.
- 2103 to be consistant with other fetch, add a return before `{`

sirmspencer:
Line 2051, 2052, I think you miss understood me. You can keep the confirmation-quit div as it was instead of the way you did it with <Fragment>.  z-index is a css property.  Add z-index: 1000 to confirmation-quit, and z-index: 1001 to modal-dialog

PR #534, opened 2020-06-18
remove latest image mosaic from the securewatch dropdowns

biplovbhandari:


lambdatronic:
collection.js:
- Lines 253-257: There is no need for a DEFAULT option in the
  dropdown. Also, directly mutating the values of DOM elements is not
  correct React programming style. Instead, use this code for the
  callback body:
```js
this.updateSecureWatchSingleLayer(dom.getElementById("securewatch-option-select"))
```

- Line 449: Call `resetSecureWatchLayer` instead of `this.setState` for consistency when switching imagery types and .

- Line 1753: Remove the DEFAULT option entry. We want the first
  available date to be selected when a new plot is loaded. If there
  are no dates, the dropdown should be empty.

sirmspencer:
I think all of these UI changes can be much simpler.  

To expand on / reinforce Garys comment, you should almost never need to access the DOM directly in react. Conditionally render each component based on the state. So instead of setting a value to the DOM when you are setting the state in getSecureWatchAvailableDates, conditionally render that option in `secureWatchMenus` where securewatch-option-select is rendered. 

In `secureWatchMenus` Do not have a DEFAULT option.  Create a conditional statement to show a disabled select with the single option "No Layers Available" if the list of available dates is 0 in length. Otherwise show the select as before, just without the "Latest Image Mosaic" option. 

I do not think you need a new openlayers route. In updateSecureWatchSingleLayer just remove the layer if there is no valid featureId.

sirmspencer:
- 108, 449, 556, resetSecureWatchLayer is not needed, you already set the state in getSecureWatchAvailableDates.  These two function are always called together.
- 253 - 259, If the conditional creates a side affect, use a full if () statement.
- 254, 564 you do not need a separate function, just call updateSecureWatchSingleLayer(null, "", "")
- 576, check for null featureId and show "No available layers"
- 587, remove extra line return
- 1760, "No available layers"

PR #537, opened 2020-06-21
load plot geometry in the gee script

biplovbhandari:


sirmspencer:
src/main/resources/public/js/mercator.js
revert all changes

- 1610 - 1614 Use the following
```
const geometry = currentPlot.geom
    ? mercator.parseGeoJson(currentPlot.geom, true)
    : mercator.getPlotPolygon(currentPlot.center,
                              currentProject.plotSize,
                              currentProject.plotShape)
```

biplovbhandari:
This won't work because this would give the coordinate in the web Mercator but we need it in WGS 84. Even if we use `reprojectToMap` to false, this still does not work because the linear units of the web Mercator and the WGS 84 do not match, as a result, calculating radius does not work. The simplest is to calculate everything in Web Mercator and convert it to WGS84 which is what `getPlotExtent` does.

biplovbhandari:
I made some changes so it reuses the `getPlotPolygon`. There are few changes in the collection to how it calls the GEE script. Those changes are made in the GEE UI as well.

PR #538, opened 2020-06-22
Proper handling of no layer result, loading state for Secure Watch

sirmspencer:


lambdatronic:
collection.js:
- Line 248: "connect key" -> "Connect ID"
- Line 256: Fix indentation.

PR #540, opened 2020-06-24
add GEE Asset

biplovbhandari:


sirmspencer:
- 165 Because you have the update button, the map should not be updating as the user is changing date options.
- 440-455,  A couple general points here.  1. I dont like to discover information that is already known. When you call this.props.setGEEImageCollectionDate(e.target), you already know if it is the start date or end date based on who calls the function.  Instead of then checking eventTarget.id, you would pass the information as a parameter. 2. If you can handle the event where the event occurs. Dont pass e.target, pass e.target.value, e.target.checked, etc.  This allows the function to stay generic and you can pass a regular value if being called somehwere else. Also, keeping the event processing where the event occurs makes it easier to look at the opject creating the event and know what the event structure is.  

Because you have the update button, the map should not be updating as the user is changing date options.  Split this into two simple functions to set start and end date.  Move the check new Date(startDate) > new Date(endDate) into updateGEEImageCollection();

- 2206 `this.props.setGEEImageCollectionStartDate(e.target.value)`
- 2017 `this.props.setGEEImageCollectionEndDate(e.target.value)`
- 618 When you build an object duplicate keys overwrite with the right most version winning.  If you spread your object you can add or midfy keys like `{ ...sourceConfig, imageVisParams: this.state.geeImageryVisParams}`
- 630 
```
{ 
	...sourceConfig,
    collectionVisParams = this.state.geeImageCollectionVisParams,
    startDate = this.state.geeImageCollectionStartDate,
    endDate = this.state.geeImageCollectionEndDate,
}
```

- 567 can you make start date max as today - 1 day?
- 724 it would be more consistant to create a new function formTextArea and check for `type === "textarea"` in form template

biplovbhandari:
For the start date, it should be fine with today. If today's image is available, it would be used for mosaic otherwise the latest available image date after today would be used in GEE.

sirmspencer:
> For the start date, it should be fine with today. If today's image is available, it would be used for mosaic otherwise the latest available image date after today would be used in GEE.

Your data check in collections and review-institution require start date < end date. In this case you can not select today for start date.

sirmspencer:
- 547 Im pretty sure getViewPolygon returns EPSG:3857, if you need EPSG:4326, use `"EPSG:3857", "EPSG:4326"` if you need EPSG:3857 use `"EPSG:3857"`

biplovbhandari:
It actually returns `EPSG:4326` calls [this](https://github.com/openforis/collect-earth-online/blob/master/src/main/resources/public/js/mercator.js#L85) to [here](https://github.com/openforis/collect-earth-online/blob/master/src/main/resources/public/js/mercator.js#L81)

sirmspencer:
> @sirmspencer It actually returns `EPSG:4326` calls [this](https://github.com/openforis/collect-earth-online/blob/master/src/main/resources/public/js/mercator.js#L85) to [here](https://github.com/openforis/collect-earth-online/blob/master/src/main/resources/public/js/mercator.js#L81)

Thank you for the reference.

Issue #479, opened 2020-06-05
Create react component for the navbar.


sirmspencer:
- [ ] The component should accept the current page components as a child with {this.props.child}
- [ ] Use react to hide and show the collapsed menu instead of bootstrap.js
- [ ] If possible, eliminate the need for "calc(100vh - 61px)" in multiple page css. (Future development)

rfontanarosa:
what do you mean with header? the navigation bar (navbar.ftl) or header.ftl?

sirmspencer:
> what do you mean with header? the navigation bar (navbar.ftl) or header.ftl?

Good question. I did mean navbar

sirmspencer:
- Create a new component file in src/main/resources/public/jsx/components called navbar.js.
- Use Create a component that will display the navbar, and then render {this.props.children}
- Pass the page component as a child like
```
<NavBar>
    <Home />
</NavBar>
```

See SideBar in collections as an example of rendering {this.props.children}

rfontanarosa:
this issue will include the convertion of the following pages into react:

- src/main/resources/template/freemarker/support.ftl
- src/main/resources/template/freemarker/register.ftl
- src/main/resources/template/freemarker/password-reset.ftl
- src/main/resources/template/freemarker/password.ftl
- src/main/resources/template/freemarker/login.ftl
- src/main/resources/template/freemarker/geo-dash-navbar.ftl
- src/main/resources/template/freemarker/about.ftl

PR #545, opened 2020-06-29
Login Page from Freemarker to React

rfontanarosa:
Login page has been converted. #479


rfontanarosa:
Hi @sirmspencer, I've converted the Login page into React. I have some considerations about it:

- flash_messenger will be deprecated, I've used a JS alert but, in my opinion, this should replaced by a modal.
- returnurl and the redirects are now managed from front-end instead of back-end
- we can improve login, register, password and password-reset adding a better (and shared) form validation
- I've kept the same query parameters structure to not change the back-end code too much, however this can be replaced sending the parameters via json
- I had to change (also in JsonUser.java) the return type. I hope that it won't be a problem.

sirmspencer:
src/main/resources/public/jsx/login.js
- You can drop documentRoot in the entire page
- 13 this is a duplicate functon to collection.js.  Move this into textUtils and import the function.
- 17 submit -> requestLogin
- 17 / 18 you dont need the event or e.preventDefault(); if you move the function from onSubmit to onClick for the button
- 34 In the else (:) portion of your conditional, this.props.returnurl would = "" so there is no reason to include it.  Also if you have the chance to use a positive comparitor, its nicer. `const redirectUrl = this.props.returnurl === "" ? "/home" : this.props.returnurl;`
- 35 after you drop document root, this and 34 can fit on one line without a variable.`window.location = this.props.returnurl === "" ? "/home" : this.props.returnurl;`
- 47 add a return between the two functions.
- setEmail, setPassword, Handle the event where it happens.  47 `setEmail = (newEmail) => this.setState({ email: newEmail });`, 59 `{e => this.setEmail(e.target.value)}`
- 56 drop the form tag and move the submit function to the button onclick event
- 71 card-footer is not defined.

rfontanarosa:
I've quickly done everything except the removal of tag form and the definition of card-footer.

I've left the tag form to keep the functionality which allows to press enter for sending it. Otherwise we have to call a onkeyup function on every input field and check if the event.which is 13.

card-footer is defined into bootstrap.css: it defines just the top-border.

sirmspencer:
src/main/resources/public/jsx/login.js
- 14, 15, if you are leaving the form action, handle the event were it occurs. `requestLogin = (e) => {` and on 53 `<form onSubmit={(e) {e.preventDefault(); this.requestLogin();}}>`

src/main/resources/public/css/custom.css
481, .card and .card-header are bootstrap classes.  Do not overwrite them in a way that will affect the rest of the site.  This can cause confusion. For buttons, there is the bootstrap class `btn` we use, then the helper `btn-lightgreen` in the custom css.  Do something similar here with helper classes `card-lightgreen`, `card-header-bg-lightgreen`, etc.

src/main/resources/template/freemarker/login.ftl
- 2 I was able to get the PR merged already to move the header into react.  If you merge master into this PR, you should be able to also use the react component header instead of the ftl header.

sirmspencer:
src/main/resources/public/jsx/login.js
- 20 remove console.log
- 33 is unreachable, you can remove this return.
- 55, 62, 66 Do not use () around a single param `e =>`

src/main/resources/public/css/custom.css

- 481, This method is not as clear as the other options I mentioned.  It still leaves known bootstrap classes on the page with non default behavior.  There are two options that are easier to follow.
   - Copy the css you want from from bootstrap and make new uniquely named css classes.
   - Make helper css classes like we do with btn and btn-lightgreen (in this case btn is bootstrap, and btn-lightgreen is our helper function). 

collection.js
- 669 update this to use getQueryString from textUtils.js

PR #554, opened 2020-07-01
create/update project visibility change options

biplovbhandari:


sirmspencer:
- 126, 210 you added the * to all imagery instead of just institution imagery. However looking at the UI I think you can remove it here completely and move it to the "Institution Private Imagery" header (so just one * total)
- 221 Add the * here "Institution Private Imagery*"
- 227 "* Institution imagery will only be visible to institution members, no matter the project privacy level."

- 170 a single duplicate is ok here. `const projectImages = [...projectImageryList, projectDetails.imageryId]` 
- 193 "Projects with privacy level of " + projectDetails.privacyLevel + " require at least one public imagery."
- 191-194 !some = every, use the positive version when possible.  Also I swithed the order of your comparison, its easer to follow what we are checking for.  
```
} else if (["public", "users"].includes(projectDetails.privacyLevel)
            && [...projectImageryList, projectDetails.imageryId]
                .every(id => imageryList.some(il => il.id === id && il.visibility === "private"))) {
    alert("Projects with privacy level of " + projectDetails.privacyLevel + " require at least one public imagery.");
    return false;
 ```

- 95-104 Same changes as create-project

PR #555, opened 2020-07-01
Convert remaining ftl pages

sirmspencer:


rfontanarosa:
I've fixed "page not found" page. The other pages are working properly.

Maybe we can add cliend-side form validation here. There are two possible ways: adding code (more flexible), use Bootstrap (based on HTML 5) or use native HTML 5 (it is already used by the email fields). Using HTML 5 validation means adding "required" attribute and "pattern" attribute on the fields, but we can't check matching fields or easy customize error messages (they are also translated by the browser, I see them in Italian).

rfontanarosa:
email_taken sql function is used twice. probably the second parameter (userId) is not needed.

PR #558, opened 2020-07-02
copy imagery from project templates

biplovbhandari:


sirmspencer:
- 272 we are phasing out doucmentRoot.  Do not use it in this new function.
- 277 since you are using the id in the .filter, it makes sense to do the .map first.
```
projectImageryList: data
	.map(imagery => imagery.id)
    .filter(imageryId => institutionImageryIds.includes(imageryId)),
```
- 284 I think we should refrain from using an alert here, this action happens too often and could result in repeat alerts.  Set projectImageryList to [] when the api fails

sirmspencer:
- 279 .filter(imageryId => institutionImageryIds.includes(imageryId)),

PR #560, opened 2020-07-07
Add some buffer to random distributions.

### LINKED ISSUE #64: Plots with random dots - dots too close together
In the plots where the dots are not systematic but random, each dot should have a buffer area (a couple of meters?) so that no other dot can be placed there. If the dots are very close together it makes the classification very complicated( selecting the right dot) and also looks bad.

Check the code from the Collect Earth circle plot generator which uses random positioning of the dots : https://github.com/openforis/collect-earth/blob/master/collect-earth/collect-earth-sampler/src/main/java/org/openforis/collect/earth/sampler/processor/CircleKmlGenerator.java

sirmspencer:
Fixes #64

### DIFF
--- a/src/main/java/org/openforis/ceo/utils/ProjectUtils.java
+++ b/src/main/java/org/openforis/ceo/utils/ProjectUtils.java
@@ -302,7 +302,7 @@ public class ProjectUtils {
         var yRange = top - bottom;
         var buffer = xRange / 50;
         return Stream.generate(() -> new Double[]{left + Math.floor(Math.random() * xRange / buffer) * buffer + (buffer / 2),
-                bottom + Math.floor(Math.random() * yRange / buffer) * buffer + (buffer / 2)})
+                                     bottom + Math.floor(Math.random() * yRange / buffer) * buffer + (buffer / 2)})
                 .limit(numPoints)
                 .map(point -> reprojectPoint(point, 3857, 4326))
                 .toArray(Double[][]::new);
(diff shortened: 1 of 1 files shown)


sirmspencer:
More buffer can be added, but then we need to start looking at realistic limit of points in the shape.

lambdatronic:
ProjectUtils.java:
- Lines 303-305: This is clearer and more efficient:
```java
var buffer = xRange / 100;
return Stream.generate(() -> new Double[]{left + buffer + Math.random() * (xRange - buffer * 2),
                                          bottom + buffer + Math.random() * (yRange - buffer * 2)});
```
- Lines 346-351: Revert back to my more efficient algorithm. The code below includes the same buffer offset as is used in createRandomPointsInBounds (i.e., 1% of the plotSize):
```java
return Stream.generate(() -> 2.0 * Math.PI * Math.random())
        .limit(samplesPerPlot)
        .map(offsetAngle -> {
            var offsetMagnitude = (radius - buffer) * Math.random();
            var xOffset = offsetMagnitude * Math.cos(offsetAngle);
            var yOffset = offsetMagnitude * Math.sin(offsetAngle);
            return reprojectPoint(new Double[]{plotX + xOffset, plotY + yOffset}, 3857, 4326);
        })
        .toArray(Double[][]::new);
```
By comparison, this algorithm uses these ops per point generated:
```
2 +
5 *
1 -
2 Math.random
1 Math.cos
1 Math.sin
```
The algorithm you presented uses these ops per point generated:
```
3 +
3 -
4 *
3 /
1 <
2 Math.random
2 Math.floor
1 Math.sqrt
2 Math.pow
```
You also generate random points that will be thrown away by the filter if they are out of bounds.
Doing calculations with polar coordinates is often more efficient when working with circles.

sirmspencer:
- Lines 346-351: Revert back to my more efficient algorithm. The code below includes the same buffer offset as is used in createRandomPointsInBounds (i.e., 1% of the plotSize):
- Doing calculations with polar coordinates is often more efficient when working with circles.
    - this might be true for efficiency, but what happens is that the points end up being more clustered to the center of the circle.

PR #565, opened 2020-07-08
move collection menu controls to separate components

biplovbhandari:


sirmspencer:
- 230-232 You can just call this.setImageryAttribution("") before setting the state.  That way any source that does not have a control will use the default attribution.
- 787 do this in the PlanetDaily component.  
- 1001, 1412 highlightSamplesByQuestion is not used
- 1413 selectedQuestion is not used
- 1397, 1407, 1430, 1440, 1450, 1460 Dont rename variables when passing as props. Use `baseMapSource={props.baseMapSource}`.  It might make more sense to rename the prop in ImageryOptions, line 996 `currentImageryId={this.state.currentImagery.id}`
- 1393-1465 I think you should be able to create an object `commonProps` once with all the common props, and then spread it into each compont.   
```
<GEEImageMenus ...commonProps />
```

src/main/resources/public/jsx/imagery/collectionMenuControls.js
- PlanetMenus there is some redundancy in this component. The way we did updates using componentDidUpdate in collection.js was really meant to simplify all of our async process (init map, get imagery, get project information, get plot, etc).  Its not really needed here because there are no fetch calls.  You can reduce your lifecycle and set functions to 
```
componentDidMount = () => {
    this.setState({
        imageryYearPlanet: parseInt(this.props.sourceConfig.year),
        imageryMonthPlanet: parseInt(this.props.sourceConfig.month),
    }, () => this.updatePlanetLayer());
}

setStateAndUpdate = (key, newValue) => {
    this.setState({ [key]: newValue }, () => this.updatePlanetLayer());
}

updatePlanetLayer = () => {
    const { imageryMonthPlanet, imageryYearPlanet } = this.state;
    this.props.setImageryAttribution(" | Monthly Mosaic of " + imageryYearPlanet +
                                     ", " + monthlyMapping[imageryMonthPlanet]);
    this.props.setImageryAttributes(this.state);
    mercator.updateLayerSource(this.props.mapConfig,
                               this.props.currentImageryId,
                               this.props.currentProjectBoundary,
                               sourceConfig => ({
                                   ...sourceConfig,
                                   month: imageryMonthPlanet < 10 ? "0" + imageryMonthPlanet : imageryMonthPlanet,
                                   year: imageryYearPlanet,
                               }),
                               this);
};
```
- PlanetDailyMenus the same thing as above.  There are three issues I see with this component that are not just coding style like above.  
    - componentDidUpdate could cause updatePlanetDailyLayer to be called twice.  It looks like you need componentDidUpdate to check if the plot changed.  But the other conditional can be reworked like above  
    - Your set function update the attribution but not the layer.  The attribution and attributes should be set in updatePlanetDailyLayer after you validate that you are going to update the layer. 
    - I think there is a bug here with the planet-layer-switcher.  It is added when any layer is Planet on map initialization.  It is then removed in several places, and not added back when switching to a Planet layer.  It should not be added in createSource (mercator.js).  It should be added every time a planet layer is selected (or is default).

- SentinelMenus, this can be reworked like PlanetMenus.  Why do you need separate state for Sentinal1 and Sentinal2?  I dont think a source can be both.

- GEEImageMenus You are updating attributes before the user clicks update.  Move updateImageryAttributes into updateGEEImagery.  geeImageryVisParams should be in the attribution.

- GEEImageCollectionMenus You are updating attributes before the user clicks update.  Move updateImageryAttributes into updateGEEImagery.  geeImageCollectionVisParams should be in the attribution.

For all of these components, you can shorten the state names.  `imageryYearPlanet` -> `year` for example.  This should help reduce the amount of text that is output in the imageryAttribution JSON block in the raw results.

sirmspencer:
- I see your point about not needed to update the layer source onMount.  19-21 and 36-37 are doing the same thing.  Update `updateImageryAttributes` to `updateImageryInformation` and update the attributes and attribution together.
- 331-333, 350-351 Same as above
- 439-440, 450-451 Same as above
- 510-512, 528, Should you be updating the attributes at the same time as the attribution?  I think this should be following the pattern as above.

- 14, 175, 504 do not use an arrow function for lifecycle functions
- 9, 10, 16, 17, 60, 72, 319, 320, 327, 328, 393, 405 It does not make sense to define these as integers. The HTML range input returns a string.  You use the value as a string more than an integer.  This would be much simpler if you updated monthlyMapping to take a string instead of a string.
- 557, 570, 580, these fit on one line now that you are not calling a function after
- In general, Gary wants semi colons after our arrow functions, even though its not required, and our linter does not flag it.  Can you go back through and add them where they are missing.

sirmspencer:
- 532 be consistant. Either destructure all of the variables or use this.state statements.

sirmspencer:
the last change we need to make is to allow created components to be hidden (so they maintain state), and then create one for each imagery option.  Then as the imagery source is change, update which imagery has visible={true} to show the corresponding controller.

sirmspencer:
- 112 This is redundant `updatePlanetDailyLayer` already calls `updateImageryInformation`
- 117 This is then redundant becuase it will only be called from inside `updatePlanetDailyLayer`
- 397, 503, 585 Does this function need to check for this.props.visible like PlanetDailyMenus since you call the same `mercator.updateLayerSource`?  In PlanetDailyMenus you also check props.visible in didMount, which is redundant.
- Please verify that all of your components are doing the same thing.

Issue #444, opened 2020-04-03
Letters allowed in numeric question answers


kdyson:
Question types (button-number, etc.) allow text (non-numeric characters) in answers.

Example:
![image](https://user-images.githubusercontent.com/6980290/78412534-8b590f00-75c8-11ea-9810-79129566300f.png)

PR #566, opened 2020-07-08
survey answer text validation

rfontanarosa:
This solves #444

### DIFF
--- a/src/main/resources/public/jsx/components/SurveyDesign.js
+++ b/src/main/resources/public/jsx/components/SurveyDesign.js
@@ -487,17 +487,8 @@ class NewAnswerDesigner extends React.Component {
         }
     };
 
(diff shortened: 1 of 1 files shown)


sirmspencer:
- 494, since you are checking as they type, do not throw and alert.  Repeat alerts are quite annoying.

rfontanarosa:
it's time to think about form validation. :)

sirmspencer:
> it's time to think about form validation. :)

Sure, and you can easily move your regex check to when the user clicks save, instead of as they type.

sirmspencer:
We may need to update your regex, it does not allow negatives or decimals.  I think the easiest solution is to just change the input type to "number" on line 517, then you dont need any regex validation.

Issue #569, opened 2020-07-08
Create an example solution using form validation


sirmspencer:
Chose a smaller page like login or register and show what you would like to add for form validation.

sirmspencer:
This is very low priority at the moment.

sirmspencer:
This is low priority right now.

PR #570, opened 2020-07-09
Login page with form validation

rfontanarosa:
This is an incomplete DEMO for #569 

**Email REGEX**
- server side regex emailPattern is incompatible with Javascript and I didn't find a proper way to use it (https://stackoverflow.com/questions/24951638/javascript-regular-expression-iinf)
- To validate email I've used a regep found here: https://emailregex.com/. Probably the server side regex and the client side should be aligned.

**HTML 5 VALIDATION**
- Hard to control, the messages are automatically translated by the browser, is based on CSS pseudo-classes

**REACT VALIDATION**
- First thing to to is disable HTML 5 validation adding noValidate attribute to form tag
- CSS pseudo-classes are impossible to manage from React, the only way is check the state
- We can validate the entire form when submitted or single input field when an input loses focus (using onBlur event). I've implemented just the first one
- We can also control the form validation status using React state, this way we can enable/disable submit button
- The state variables which contain the error message are strings, we can change them into array to support multiple errors


rfontanarosa:
HTML 5 FORM VALIDATION PRO AND CONS

PRO
- It's built-in and implemented by all modern browsers so it doesn't requires additional frameworks
- Easy to implement and very effective
- Customizable using Javascript API (https://css-tricks.com/form-validation-part-2-constraint-validation-api-javascript/)

CONS
- These are all the possible validators: https://developer.mozilla.org/en-US/docs/Web/API/ValidityState. If we need something that depends on another input (like password and confirm password) we have to write code.
- By default it shows only the first error message, showing all the errors means write code (https://www.tjvantoll.com/2012/08/05/html5-form-validation-showing-all-error-messages/)
- By default the errors are shown in a popup, changing this means write code (https://stackoverflow.com/questions/10753881/changing-the-language-of-error-message-in-required-field-in-html5-contact-form)
- By default it shows translated and predefined error messages, define custom error messages means write code (https://www.geeksforgeeks.org/form-required-attribute-with-a-custom-validation-message-in-html5/)

sirmspencer:
I like the look of the fields when there is an error.  However this not quite yet scale-able.  There are many ways to make this more scale-able, here are a couple suggestions to start.
- Create a component for the input box, and check validity inside the component by type. Right now you check by name of the field, and all the validation code would have to be copied everywhere the component is used.  The input component would only need to take the state value, an on change for the sate and toggle error.
- On the log in page, keep the form state the same (ie password: email:) and have one state object for the errors, it can start empty.  `toggleError = (key) => this.setState({ [key]: !this.state.errors[key] })` Then when you check the errors on submit you can do `Object.values(this.state.errors).every(e => e)`

sirmspencer:
autocomplete off has never made sense to me. I think we should omit autocomplete = "off".  If you want to get fancy you could add informational autocomplete per https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

sirmspencer:
This PR is stale.  When the code is updated, create a new PR.

PR #582, opened 2020-07-15
FB add https

sirmspencer:


lambdatronic:
deps.edn:
- Line 14: pyregence.https -> org.openforis.ceo.https

https.clj:
- Line 1: org.openforis.ceo.handler -> org.openforis.ceo.https
- Line 5: pyregence.logging -> org.openforis.ceo.logging
- Line 34: Move `log-str` call to next line.
- Line 56: `[certbot path root]` -> `[certbot-path-root]`. Otherwise, this seems to indicate three args.

handler.clj:
- Please update production-app to use wrap-ssl-redirect and both
  production-app and development-app to use wrap-common as in
  pyregence.handler.

PR #586, opened 2020-07-15
default value and new style for unsubscribe mailing list page

### LINKED ISSUE #578: Update unsubscribe page
- [x] If the user is already logged in, fill the box with their email.

- [x] The form is missing the same styling that you used it the updated the log in page.

rfontanarosa:
fixes #578


sirmspencer:
src/main/resources/public/jsx/unsubscribe-mailing-list.js
- 9 If you are going to initialize the state to a prop, do so in componentDidMount.  It is much clearer that you are intending for it to be set once and not remain the same value as the prop.

PR #587, opened 2020-07-15
missing institution logo fix

### LINKED ISSUE #579: Show ceo logo if there is no logo file for an institution
This might work https://stackoverflow.com/questions/3984287/how-to-show-alternate-image-if-source-image-is-not-found-onerror-working-in-ie

rfontanarosa:
fixes #579

### DIFF
--- a/src/main/resources/public/jsx/components/PageComponents.js
+++ b/src/main/resources/public/jsx/components/PageComponents.js
@@ -240,3 +240,40 @@ export function LogoBanner() {
         </div>
     );
 }
+
+export class SafeImage extends React.Component {
+    constructor(props) {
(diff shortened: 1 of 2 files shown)


sirmspencer:
I like where you are going with this.  I think it can be simplified:
src/main/resources/public/jsx/components/PageComponents.js
- 248, there is no reason to set the state to the src when its already a prop. 
- 255, if props.src changes, just set the error back to false
- 269, const { src, fallbackSrc, ...extraProps } = this.props;
- 273, onError={this.setState({ error: true })}
- 272, src={this.state.error ? fallbackSrc : src}
- 274, {...extraProps}

PR #589, opened 2020-07-16
new map toggle button

### LINKED ISSUE #568: Make the bar to hide the project list panel easier to see.
- Make the button wider
- Make the button short and centered vertically.  If it is only 2rem tall, it will stick out like a tab instead of blend in.

rfontanarosa:
fixes #568

### DIFF
--- a/src/main/resources/public/css/custom.css
+++ b/src/main/resources/public/css/custom.css
@@ -55,11 +55,6 @@
     height: calc(100vh - 60px);
 }
 
-.ceo-map-toggle{
-    position:fixed;
-    height: calc(100vh - 60px);
-}
-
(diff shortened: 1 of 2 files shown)


sirmspencer:
The new button looks great.  Can you change the cursor to pointer when the mouse is over the button?

rfontanarosa:
what we can do is move the zoom controls 2rem to the right and add our toggle button not in the middle but just to the left of zoom controls.

sirmspencer:
> what we can do is move the zoom controls 2rem to the right and add our toggle button not in the middle but just to the left of zoom controls.

The complaint has been that its too hard to find.  Id rather leave it in the middle.

PR #590, opened 2020-07-16
added tab descriptions

### LINKED ISSUE #580: Add descriptions for each tab on the institution review page
- Imagery: "This is a list of available imagery for this institution.  For each project you can select to use some or all of these imagery"
- Projects: "This is a list of all institution projects.  The color around the name shows its progress.  Red indicates that it has no plots collected, yellow indicates that some plots have been collected, and green indicates that all plots have been selected"
- Users:  "This is a list of all institution users.  A institution admin can create and update projects and imagery for the institution. Members can view projects with the visibility Institution or higher"

rfontanarosa:
fixes #580

### DIFF
--- a/src/main/resources/public/jsx/review-institution.js
+++ b/src/main/resources/public/jsx/review-institution.js
@@ -12,7 +12,7 @@ class ReviewInstitution extends React.Component {
         this.state = {
             imageryCount: 0,
             usersCount: 0,
-            projectList: [],
+            projectList: null,
             isAdmin: false,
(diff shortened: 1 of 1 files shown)


sirmspencer:
- 75, check for null inside headerTab instead of the function call. On line 53 {count || 0}

rfontanarosa:
the problem is that I have to check if state.projectList is not null, otherwise accessing on an attribute of null raises an error:
`this.state.projectList ? this.state.projectList.length : 0`

PR #597, opened 2020-07-17
FB Converted PostgresUsers.java into users.clj

lambdatronic:
----

#


sirmspencer:
- 53 fix column alignment
- 89 I think you can move this check to get-update-account-errors.  If the email is the same as the current email, it returns false.
- 106-110, if we check "email_taken" in get-update-account-errors, we can create one "update_user_information" sql function.  Maybe not in the PR, but a TODO note would help remember.
- 93 this will replace the entire session with just userName
- 114 This is inconsistant with other functions which deref request in the function parameter.  Can we just deref the requests before these functions are all called?
- 143 can you move `user (first (call-sql "get_user" email))` into the first let, and then have get-reset-password-errors process all the possible errors at once?
- 156 We wont need a map if the SQL is updated to not return reset-key and return text role.
- 159, 242 This is inconsistant with other functions which deref request in the function parameter.
- 165 I dont think we should be passing the reset key to the front end. 
- 187 This is not needed.  It was a stub for a matching JSON version that was implimented.
- 191 update-institution-role looks translated correctly, but this logic is quite odd.  
- 192 I think this is another spot where the query param needs to be updated to accout-id to distinguish params.
- 218, 224 (get-institution-by-id institution-id) is not used, you can just return (data-response "")
- 232 - 239 can you use cond here or create a get-mailing-list-errors function?

sirmspencer:
src/clj/org/openforis/ceo/utils/mail.clj
- Do we need another file? There is already email.clj

email-server.edn
- This is used in email.clj. 

src/clj/org/openforis/ceo/email.clj
- 6 You deleted email-server.edn, update to mail-config.edn

lambdatronic:
users.clj:
- 157 I dont think we should be passing the reset key to the front
  end. We wont need a map if the SQL is updated to not return
  reset-key and return text role.

  The reset key is not returned in the response from get-all-users.

  I added a FIXME to capture your thoughts for later.

- 166 I dont think we should be passing the reset key to the front
  end.

  We shouldn't modify the API behavior during this rewrite as it can
  quickly lead to a big list of changes we need to make on the
  front-end to support said changes. After the rewrite has been merged
  and deployed, we can improve upon the existing behavior in future
  PRs that modify the front-end and back-end together.

  I added a FIXME to capture your thoughts for later.

- 189 update-institution-role looks translated correctly, but this
  logic is quite odd. I can look at it when we start testing.

  Sounds good. There is a FIXME already on the function for further improvements.

- 190 I think this is another spot where the query param needs to be
  updated to accout-id to distinguish params.

  I added a FIXME.

PR #605, opened 2020-07-22
FB convert PostgresImagery.java to imagery.clj

sirmspencer:


lambdatronic:
imagery.clj:
- Lines 13,27,34,38: Non-idiomatic function name. Sequence functions
  take their sequence as the last argument (to be compatible with
  `->>` semantics) and should take a function to apply to the elements
  in the sequence. Consider renaming this to `prepare-imagery` to
  match `prepare-institution` in institutions.clj.
- Line 20: `(json/read-str extent)`
- Line 21: `(if admin? (json/read-str source_config) (clean-source (json/read-str source_config)))`
- Line 33: :token-key -> :tokenKey (camelCase matches the Java
  implementation and is consistent with all other params and sessions
  keywords (except flash-message, which should be renamed to
  camelCase).
- Lines 42-45: Incorrect implementation. `get-imagery-source-config`
  returns a source-config map, not a response map. Also, `call-sql`
  can only call functions, not arbitrary SQL. My comments for
  imagery.sql include a new function definition to resolve this issue. Use this code instead:
```clojure
(defn get-imagery-source-config [imagery-id]
  (if-let [imagery (first (call-sql "select_imagery_by_id" imagery-id))]
    (json/read-str (:source_config imagery))
    {}))
```
- Line 51: No `json/read-str` since we have to pass a stringified version of the JSON to "add_institution_imagery".
- Lines 52,93: `str-bool` -> `str->bool`
- Lines 52,93: `str->bool` will not default to true if :addToAllProjects
  is missing because `(boolean nil)` is `false` and will not throw an
  exception. Use this code instead:
```clojure
(tc/str->bool (:addToAllProjects params "true"))
```
- Line 72: :geeParam -> :geeParams
- Line 85: `(json/write-str source-config)`
- Line 92: No `json/read-str` since we have to pass a stringified version of the JSON to "update_institution_imagery".
- Line 94: `call-sql` can only call functions, not arbitrary SQL. My
  comments for imagery.sql include a new function definition to
  resolve this issue. Use this code instead:
```clojure
(:institution_rid (first (call-sql "select_imagery_by_id" imagery-id)))
```

type_conversion.clj:
- Lines 9,17: `(catch Exception _ default)`

imagery.sql:
- Add this function:
```sql
CREATE OR REPLACE FUNCTION select_imagery_by_id(_imagery_id integer)
 RETURNS setOf imagery_return AS $$

    SELECT imagery_uid, institution_rid, visibility, title, attribution, extent, source_config
    FROM imagery
    WHERE imagery_uid = _imagery_id

$$ LANGUAGE SQL;
```

lambdatronic:
imagery.clj:
- Lines 14,20: Use `tc/jsonb->clj`
- Lines 33,34: tokenKey -> token-key (Clojure variables use kebab-case. You just needed to change the :tokenKey params keyword to camelCase.)

type_conversion.clj:
- Add these function aliases for completeness:
```clojure
(def clj->json json/write-str)

(def jsonb->json str)
```
- Line 18: `(str string)` -> `string` (If string isn't a string, the function has been called incorrectly and should throw an error.
           Also, add back `try/catch` to match `str->int`'s behavior. Otherwise, `(str->bool 10 true)` returns false, which is probably not what we intend.
           The following code ensures that `str->int` and `str->bool` both have these properties:
           1. If string = nil, return default
           2. If string is not a string, return default
           3. If string is an incorrectly formed string, return default
           4. Otherwise, return the parsed value
```clojure
(defn str->bool
  ([string]
   (str->bool string false))
  ([string default]
   (if string
     (try
       (Boolean/parseBoolean string)
       (catch Exception _ default))
     default)))
```

PR #611, opened 2020-07-24
imagery delete project btn

### LINKED ISSUE #601: Add delete button for projects on the review institution page
Add next to edit, and make available only for institution admins like the edit button

rfontanarosa:
fixes #601


sirmspencer:
- 46 alert after you update the list
- 47 please call getProjectList after deleting.  This is what I call single source of truth. We think the project was deleted but we dont really know until we call the server to check.
- 372 revert this change
- 978 what is the default event that you are trying to prevent?

PR #612, opened 2020-07-24
add securewatch base url in form

biplovbhandari:


sirmspencer:
src/main/java/org/openforis/ceo/Proxy.java
- 122 the base source is not a secret, is it already available to you in getSecureWatchAvailableDates in collection.js?  This would eliminate the need for a back end change.

PR #616, opened 2020-07-28
FB Converted PostgresInstitutions.java to institutions.clj

lambdatronic:
----

#


sirmspencer:
- 32 This is a legacy way to respond, add TODO note to return an errror instead of an empty institution map.
- 32 I think this can be private (after updating users.clj to not return the unused institution object)
- 68 If this path is an error return and error text "Error creating institution."
- 87, 93 We dont use institution-id, you can return ""
- 88 If this path is an error return and error text "Institution not found."

lambdatronic:
- 68 If this path is an error return and error text "Error creating
  institution."

  This deviates from the Java behavior. I added a FIXME to make your
  requested change and update the front-end code at the same time.

- 89 If this path is an error return and error text "Institution not
  found."

  This deviates from the Java behavior. I added a FIXME to make your
  requested change and update the front-end code at the same time.

PR #617, opened 2020-07-28
show total plots and samples per plot

biplovbhandari:


sirmspencer:
It does not seem to be calculating plots or samples when gridded is selected.
Can you make the font size a little bigger?

sirmspencer:
- 950 You cant call a non pure function in a functional component (or render function).  This is causing an infinite loop.  When you call checkPlotSampleLimitError it updates the same values that you are rendering, which causes a re-render. You will probably need to keep the sample limit object in state somewhere and update it on change for the inputs that affect the calculation.

sirmspencer:
- 70 why do you need to check limits on mount?  the user has not had time to enter any numbers to calculate
- 337 call this.checkPlotSampleLimitError() like you do for templateId === 0
- 372 only recalculate if the key is one that affects the total.  We dont need to be recalculating as the user is typing a project name, description, etc. 
 
This might be a good case for componentDidUpdate.  You can eliminate all of these checkPlotSampleLimitError if you watch for the few variables that affect the calculation in componentDidUpdate.

- 500-506 I think this math simplifies a little
```
const buffer = Number(projectDetails.plotSize);
const xRange = upperRight[0] - lowerLeft[0] - buffer;
const yRange = upperRight[1] - lowerLeft[1] - buffer;
```
- 574 You do not need to pass checkPlotSampleLimitError

PR #625, opened 2020-07-28
add-geodash-imagery route has been removed

### LINKED ISSUE #604: Update the widget layout editor to use add-institution-imagery instead of add-geodash-imagery
- [x] Build the source config on the front end
- [x] Ask Billy if addToAllProjects should be true or false.  My guess is false.

rfontanarosa:
fixes #604


billyz313:
Please also remove the option to add institution imagery from the widget layout editor as this is no longer an option.

rfontanarosa:
> Please also remove the option to add institution imagery from the widget layout editor as this is no longer an option.

removing "Add Asset to institution basemaps" the state variable "addCustomImagery" will be always set to false.

this means that from geo-dash will be impossible to add an imagery at institution level. is this the correct behavior?

billyz313:
Yes, you will have to remove the checkbox as well.  It might be a good idea for the time being to replace the text that says add to institution basemaps with text that says "Adding imagery to basemaps is now available on the institution review page in the imagery tab" and maybe even including a link to their institution page (open in new tab so we don't disrupt the users widget build)

sirmspencer:
- 228-248 we changed this for the API call.  Since we are not making the API, id suggest reverting the code so there is less change that we broke something else.
- 1111, looking good, can you make the link just around `institution review page` so its more obvious that its a link and not just blue text?

PR #629, opened 2020-07-29
FB Basic routing for view pages

sirmspencer:


lambdatronic:
handler.clj:
- Line 31: /accout -> /account
- Line 33: /create-instituion -> /create-institution
- Line 69: Projects API (use plural for consistency)
- Line 136: Add a TODO comment about needing to implement the proxy routes

lambdatronic:
Why are we no longer implementing the /update-project-user-stats route?

sirmspencer:
Why are we no longer implementing the /update-project-user-stats route?
- See my comment for users.clj.  This route is an old JSON database route.

PR #633, opened 2020-07-30
show securewatch message at the project level

biplovbhandari:


sirmspencer:
src/main/resources/public/img/securewatch-go-to-plot.png
- can you put some white space horizontally around your text? 

src/main/resources/public/jsx/imagery/collectionMenuControls.js
- 205 It looks like its working OK without this check. Since secure watch is the only one that uses updateLayerWmsParams, its not needed. But if you want to make this future proof, keep, but nest under `if (prevProps.visible !== visible)` so it only fires once, when the user changes back to the securewatch layer.

PR #634, opened 2020-07-30
flagged plot

### LINKED ISSUE #621: Show that a plot is flagged
Right now it looks unanswered

rfontanarosa:
fixes #621


sirmspencer:
src/main/resources/public/jsx/components/SurveyCollection.js
- 64, you dont pass isFlagged.  I think this.props.isFlagged is visible here.
- 141, Now that you hide the question tree to show that it is flagged, you need a way to unflag the plot and answer the questions.  I think if you add a clear button next to flagged to show the questions again that would work.  In this case, it might make sense to also hide the questions when clicking flag plot.  Then dont set the plot as flagged until the user hits save.

sirmspencer:
A new route is overkill here.  
- Clicking unflag should just show the questions again.  
- When a user hits save, if the plot is flagged, call flagPlotInDB, if the questions are answered, call postValuesToDB.

rfontanarosa:
I've implemented your idea. Please note that this way, to flag a plot, a user have to click twice (flag and save) instead of once.

sirmspencer:
Looks good!

collections.js
The reason your commented out state change is causing side affections is because the code is looking for changes in currentPlot
- 96 `if (this.state.currentPlot.id && this.state.currentPlot.id !== prevState.currentPlot.id) {`
- 925 
```
    toggleFlagged = () => {
        this.setState({ currentPlot: { ...this.state.currentPlot, flagged: !this.state.currentPlot.flagged }});
    };
```
- 1062 You need to enable the save button if the plot is flagged.

rfontanarosa:
you're right. I'll move these buttons above save button when the code is stable.

PR #648, opened 2020-08-05
auto-launch geo-dash settings

biplovbhandari:


sirmspencer:
- 99 do you need `this.state.currentProject.projectOptions.hasOwnProperty("autoLaunchGeoDash")`? obj.notAProperty returns undefined which is falsy.
- PostgresProject.java update default settings to include autoLaunchGeoDash and showPlotInformation

Issue #642, opened 2020-07-30
Dont allow users to remove the last user from an institution


sirmspencer:

sirmspencer:
Do after #637

sirmspencer:
Also dont allow the last admin to be removed.

PR #656, opened 2020-08-06
review institutions page list of users

biplovbhandari:


sirmspencer:
- 1086 What is the differece between `this.props.isAdmin` and `this.currentIsInstitutionAdmin()`?  It looks like this.props.isAdmin is already what you want. The information is coming from two places (isAdmin is set in InstitutionInformation).  Maybe later we can reduce the SQL calls so its only coming from one place.
- 1010-1014, I think its ok to show a member how many users there are so you can revert this code.
- 1113-1136 this code can simplify
```
{this.state.institutionUserList
    .filter(iu => iu.id === this.props.userId || this.props.isAdmin)
    .sort((a, b) => sortAlphabetically(a.email, b.email))
    .sort((a, b) => sortAlphabetically(a.institutionRole, b.institutionRole))
    .map((iu, uid) =>
        <User
            key={uid}
            user={iu}
            isAdmin={this.props.isAdmin}
            updateUserInstitutionRole={this.updateUserInstitutionRole}
        />
    )
}
```

You linked `Dont allow users to remove themselves from an institution #642`, but I dont see any changes that would prevent the user from deleting their selves

biplovbhandari:
I think the 642 is also fulfilled because a member of the institution can only be removed by the admin, ie the member cannot remove themselves from the institution.

image

PR #669, opened 2020-08-14
aoi coordinates editing

### LINKED ISSUE #639: Allow the user to edit the AOI coordinates manually
I suggest clearing the yellow boundary as soon as the user starts manipulating the input boxes.  Then show a button to show the border so each update of the form (ie each key stroke when typing the coordinates).

rfontanarosa:
fixes #639

### DIFF
--- a/src/main/resources/public/js/mercator.js
+++ b/src/main/resources/public/js/mercator.js
@@ -897,6 +897,15 @@ mercator.removeLayerById = function (mapConfig, layerId) {
     return mapConfig;
 };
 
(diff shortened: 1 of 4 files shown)


sirmspencer:
Can you add some code to handle if the user completely clears the input box?  Right now it throws an error and the page goes blank.  It might be most intuitive if you clear the boundary box for invalid boundary coordinates.  This would also be true if the values are too high. See valid_boundary for the conditions in which the boundary is valid.

Issue #674, opened 2020-08-19
Add an option to the dual imagery widget to set the default as swipe or opacity.


sirmspencer:

billyz313:
It seems that the widget UI is confusing to the user as well, so I will fix that, likely by changing the button to a toggle switch.  I will also have to add a state for default that the admin can select when setting the widget up.  To remain backward compatible I will just store Swipe_As_Default = true, if they want the new configuration otherwise nothing will be set and it will render the current way which is Opacity as default.

PR #680, opened 2020-09-01
Dual imagery widget opacity-swipe default option

billyz313:
Resolves issue #674 "Add an option to the dual imagery widget to set the default as swipe or opacity."


sirmspencer:
src/main/resources/public/css/geo-dash.css
- 172, 313, are we using kebab or camel case in our css elsewhere?

- 941 these lines are too long, break the lines like line 944.
- 929-938, these are not use more than once, you can just inline the style where the style tag is.  It may seem trivial, but it helps when I review that I dont have to chase down variables.  I almost was able to just use the web interface for Git for this review, but with the variables I had to open this branch in my editor to double check.

- 235, 327, 387, 531 not for this PR, but consider creating a `const defaultWidgetParam` outside the component instead of repeating the same initialization 4 times in the code.

PR #689, opened 2020-09-09
Review changes for non CLJ files, excluding the readme.

sirmspencer:


sirmspencer:
README.md, README.org:
- I will do this in a sperate PR.

deps.edn:
You removed the GeoTools repository and deps. Are we no longer using these?
- As far as I can tell those two repositories were just used to get the reprojection function.  I wrote our own function for now since we only needed 3857->4258 and vis versa.

review-project.js:
Line 648: You removed the documentRoot prop but left all the references to this.props.documentRoot in the file. Remove all such references.
- The remaining references are removed in master. After a rebase the two sets of changes elimate all references.

Make sure we are using response.json() everywhere now instead of response.text() to match data-response's API.
- there are no uses of response.text() in the entire code base
Make sure we aren't passing userId, userName, or userRole in any fetch requests unless we are changing them.
- Ive checked a few times. I have a note to double check one more time after the rebase.

*.java:
Delete this file.
- PR 677 removes all java files and folder.

project.sql:
Lines 936,951: Does user_plots have a plot_uid or plot_rid field? This looks incorrect.
- user_plots has a plot_rid which references plot_uid in plots.

PR #691, opened 2020-09-10
review changes for route handling

sirmspencer:


sirmspencer:
handler.clj:

Line 52: Are you sure that "referer" shouldn't be "Referer"? 
- All headers are lower cased in the request map.

Line 193: Why are we using defonce here? Is there some state associated with wrap-reload that we need to preserve?
routing.clj:
- This is tough choice, the defonce here means any changes in the handler stack are not automatically updated.  On the other hand any changes to the back end will end up causing the session to drop and therefor the user will be logged out.  In my work, the latter was the bigger issue so I think defonce is the better way to go.

routing.clj:
Line 12: TODO flatten url structure, conform name and fn, and rename geo-dash to geodash everywhere
- I need to do this after we rebase so I can coordinate with the front end.

The :user authentication just verifies that the user is logged in.  We use userId off the session now so many of the routes in question dont "need" additional security. I added it anyway so its more clear.

Line 27: Shouldn't we have :collect authentication on the [:get /geo-dash] route?
Line 45: Shouldn't we have :proj-admin authentication on the [:get /widget-layout-editor] route?
- Probably, this has been an outstanding issue. I have not gotten an answer from Billy on whether he wants to have auth on these routes. I guess we can just decide for him and add them.

Line 65: Shouldn't we have :user authentication on the [:post /logout] route?
- There is no reason the user cant access the log out route while not logged in. If we authenticate it would actually cause a loop where the user is redirected to the login page with the redirect-url set to the log-out route.

Line 149: Shouldn't we have :inst-admin authentication on the [:get /get-institution-imagery] route?
- The SQL function only returns public imagery if the user is not a member of the institution.  Adding authentication would block users from going to the SIG organization to see the public imagery.

Line 150: Shouldn't we have :collect authentication on the [:get /get-project-imagery] route?
- The SQL function only returns public imagery if the user is not in the institution. I think I can add :collect for clarity since public project will still authenticate true for the average user.

Line 152: Shouldn't we have :proj-admin authentication on the [:post /add-geodash-imagery] route?
- This route is now depreciated. It will be removed on merge with master.

PR #692, opened 2020-09-10
Review proxy.clj

sirmspencer:


sirmspencer:
Lines 16,19,39: You don't need to pass geoserver-params to this function since params already contains all of its keys.
- I dont think this is the case, geoserver-params comes from get-imagery-source-config. geoserver-params is overwriting the params from OL

lambdatronic:
proxy.clj:
- Lines 16,37: You don't need to pass geoserver-params to
  apply-default-styles since params already contains all of its keys.
  You merged them on line 36.

routing.clj:
- Drop parens around :handler functions.

PR #695, opened 2020-09-11
review changes for users

sirmspencer:


sirmspencer:
users.clj:

Line 154: Do we not use the :role attribute anywhere get-all-users is called?
Line 162: Do we not use the :role attribute anywhere get-institution-users is called?
- No, role in this context only means super user or not.  This was left over from the original version of CEO.  Role is only ever used in the session `:userRole (if (:administrator user) "admin" "user")`

Line 208: "update_institution_user_role" is defined to return an integer. Does it return a falsey value to Clojure on failure?
(sql-primitive ) returns null if no value is returned. update_institution_user_role will return null/void if no user is updated.
```
user=> (sql-primitive (call-sql "update_institution_user_role" -1 -1 ""))
09/11 09:22:35 SQL Call: SELECT * FROM update_institution_user_role(-1,-1,"")
nil
```

PR #708, opened 2020-09-17
Organize utils folder

sirmspencer:


lambdatronic:
create-project.js:
- Lines 12-14: These variables don't need to be exported.

geo-dash.js:
- Line 20: Drop space between function name and args for stylistic consistency.

generalUtils.js:
- Line 53: Drop space between function name and args for stylistic consistency.

PR #709, opened 2020-09-17
Add SvgIcon component

sirmspencer:


sirmspencer:
Im thinking now that the SvgIcon is a component not a util. However, there are too many changes to that file down the chain so I will have to move it later.

lambdatronic:
svgIcons.js:
  - Line 56: fill is not a valid style attribute for divs. Do you mean background-color?
  - Line 60: Remove blank line.

sirmspencer:
Line 56: fill is not a valid style attribute for divs. Do you mean background-color?
- It is the correct attribute to change the SVG color.  I used the same style tag in Pyregence.

PR #724, opened 2020-09-23
Convert ceoMapStyles to a function to use as a style function

### LINKED ISSUE #720: Update mercator to show sample geom when not a point

sirmspencer:
closes #720

### DIFF
--- a/src/clj/org/openforis/ceo/db/plots.clj
+++ b/src/clj/org/openforis/ceo/db/plots.clj
@@ -24,7 +24,7 @@
           (merge {:id         sample_id
                   :sampleGeom sample_geom
                   :sampleId   sampleId ;TODO I don't think we distinguish between sample_id and sampleId so this could go away
-                  :geom       (tc/jsonb->clj geom)}
+                  :geom       geom}
                  (when (< 2 (count (str value)))
                    {:value (tc/jsonb->clj value)})))
         (call-sql "select_plot_samples" plot-id project-id)))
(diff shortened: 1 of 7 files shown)


sirmspencer:
image

lambdatronic:
plots.clj:
- Lines 24-27: These changes revert fixes from PR723.

mercator.js:
- Line 784: textFillColor is not passed to this function.
- Line 785: It looks like fillColor or text can both be null here, but
  you don't explicitly define the default value to be null as you did
  in getCircleStyle. Aso, why retain the ternary fill expression on
  line 789 rather than use the explicit default value setting that you
  used on line 808?
- Line 861: mercator.getCircleStyle(radius, borderColor, borderWidth, fillColor = null)
- Line 1141: What are the possible return values for getType(). Can we get "answeredlinestring"?

sirmspencer:
Line 785: It looks like fillColor or text can both be null here, but you don't explicitly define the default value to be null as you did in getCircleStyle. Aso, why retain the ternary fill expression on line 789 rather than use the explicit default value setting that you used on line 808?
- There are no default values,  the ternary and `(text || "")` are just to keep the style function from failing hard if the call to getClusterStyle is wrong. getClusterStyle does not have the same concept of needing a filled middle with a transparent color, so there is no need to add default values. 

Line 1141: What are the possible return values for getType(). Can we get "answeredlinestring"?
- Yes, but in this PR, no. That option is gone in later PRs with a generic getGeomStyle

PR #725, opened 2020-09-23
Update webpack again for use with current workflow

sirmspencer:
Update webpack and view.clj for improved compiling.  The automated chunking by webpack is more efficient than I can figure out how to define manually.  Mainly it does a good job of splitting out vendor chunks while still chunking them by entry point (when only one entry point uses a vendor, it is a separate vendor chunk).  The automated chunking produces small chunks that incorporate too many of our entry points at once to use the file name as a reference to build the `<head>`. I update the compilation so that webpack builds a json file with all the referenced bundles / chunks.  This also allows me to add back the hash code, which is nice for eliminating over caching on file changes.


lambdatronic:
webpack.config.js
- Lines 80,90: This will throw an error if entry-points.json does not yet exist.

PR #735, opened 2020-09-24
Upgrade openlayers and other rotten packages.

sirmspencer:
There were 3 breaking changes to OL. 1) The selection tools now auto save style. Trying to do it manually doesnt work. 2) The view port for OL needs to be sized in all cases.  3) The default zoom to extent of a vector does not have any padding be default.


lambdatronic:
package.json:
- Line 81: I thought we were no longer using file-loader in webpack.config.js?

collection.js:
- Line 831: What is the purpose of this setState call?
```js
this.setState({selectedSampleId: -1});
```

mercator.js:
- Line 1029: No callback function for `select.on()`?

sirmspencer:
Line 81: I thought we were no longer using file-loader in webpack.config.js?
- Good catch, not sure what change has it added back.

Line 831: What is the purpose of this setState call?
- I believe that is un-selecting samples after the user answers.

PR #741, opened 2020-09-26
Update draw tools to create new plot samples as the user switches back to question mode.

### LINKED ISSUE #727: Allow users to edit existing samples on the plot
For example to go back and review their answers.

sirmspencer:
closes #727


lambdatronic:
collection.js:
- Line 452: this.state.mapConfig -> mapConfig
- Line 485: Missing semicolon.
- Line 463: Why does nextId count down from min rather than up from max? Are you concerned about negative nextId values?
  
PageComponents.js:
- Line 134: componentDidMount()  <= No space before parens on React lifecycle methods

mercator.js:
- Line 916: (color) -> color

sirmspencer:
Line 463: Why does nextId count down from min rather than up from max? Are you concerned about negative nextId values?
- In this PR I wanted negative numbers on purpose to show witch were the new plots.  In the next PR, I switch to max to count up.

PR #746, opened 2020-09-28
Simplify styling which now allows for any geom shape

sirmspencer:


lambdatronic:
projects.clj:
- Line 367: "shp2pgsql -s 4326 -t 2D %1 ext_tables.%2"

mercator.js:
- Line 28: RegularShape is unused.
- Lines 821-822: [Pure] Returns a style object that displays a solid point with the specified radius and fillColor.

PR #747, opened 2020-09-29
Some fixes for alternative user usage

sirmspencer:
For example, when deleting all drawn geoms or going to the next plot still in draw mode.


lambdatronic:
collection.js:
- Line 450: What is the filter expression doing here? Are we just discarding nil sampleIds? What would cause nil sampleIds?
```
const existingIds = allFeatures.map(f => f.get("sampleId")).filter(id => id);
```
- Lines 483,494: Missing semicolon.
- Line 1094: props.currentPlot && props.currentPlot.samples
- Line 1097: props.currentPlot.isFlagged => props.currentPlot.flagged
- Line 1114: props.currentPlot.isFlagged => props.currentPlot && props.currentPlot.flagged
- Line 1120: (props.currentPlot && props.currentPlot.analyses > 0)

sirmspencer:
Line 450: What is the filter expression doing here? Are we just discarding nil sampleIds? What would cause nil sampleIds?
- The new samples dont have an id assigned.

Line 1094: props.currentPlot && props.currentPlot.samples
Line 1114: props.currentPlot.isFlagged => props.currentPlot && props.currentPlot.flagged
Line 1120: (props.currentPlot && props.currentPlot.analyses > 0)
- currentPlot is now always defined. (line 25)

PR #782, opened 2020-10-20
Update type conversions to force Java type

sirmspencer:


lambdatronic:
type_conversion.clj:
- Lines 14,25,37,49,59: Cast `default` to primitive type in catch expressions.

PR #783, opened 2020-10-21
Simplify authentication functions

sirmspencer:
The two cases that will use this moving forward are: 
1. To have one route for create and review, authentication has to work on either field.
2. To use is-proj-admin? in more places without passing the entire param object


lambdatronic:
institutions.clj:
- Line 13: `call-sql` takes sql-fn as first arg.

projects.clj:
- Line 24: `call-sql` takes sql-fn as first arg.

PR #784, opened 2020-10-21
Add loading / saving / fetching plot message on collection page

sirmspencer:
image

closes #321 

It looks like when we moved the imagery controls into their own components, the use of the spinner for planet daily was lost. However I didnt like where it was anyway.  If we want a loading message for Planet, lets rethink where it should go.

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -597,7 +597,6 @@ ul.samplevalue {
     height: 4rem;
     border-radius: 2rem;
     border-bottom: 0.5rem solid #3fabc6;
-    visibility: hidden;
     animation: sweep 1s infinite linear;
     -o-animation: sweep 1s infinite linear;
     -moz-animation: sweep 1s infinite linear;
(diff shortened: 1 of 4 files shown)


lambdatronic:
collection.js:
- Line 134: Why are we wrapping a Promise.resolve() call in a new
  Promise? This seems like an unnecessary level of indirection that
  isn't needed at any of the call sites for functions using
  this.processModal:
```js
processModal = (message, callBack) =>
    this.setState({modalMessage: message},
                  () => callBack().finally(() => this.setState({modalMessage: null}))
);
```
- Lines 262-265,271-272,295-299,306-307,331-336,342-343,: Fix indentation.
- Line 529: Missing semicolon.

sirmspencer:
Line 134: Why are we wrapping a Promise.resolve() call in a new Promise?
-- Its used in postValuesToDB to nest messages. Otherwise this.setState() get the new message and the null from the last message in the same update cycle.

PR #787, opened 2020-10-23
Allow admins to collect in unpublished mode and clear samples when publishing

sirmspencer:


lambdatronic:
projects.clj:
- Lines 462,498,561,590,638,652,654,: allow-drawn-samples -> allow-drawn-samples?

project.sql:
- Lines 1642-1643: `lon` and `lat` are stored as double precision values.

PR #800, opened 2020-10-28
Allow edit survey questions

sirmspencer:


lambdatronic:
projects.clj:
- Line 667: `(:imageryId params nil)` -> `(:imageryId params)` This is
  redundant. If :imageryId is not in params, the implicit `get`
  already returns `nil`.
- Lines 675-676: No need to bind `original-project`. Just use the SQL call as the `if`'s test expression.
- Line 695: Only one space in " not found" string.
- Line 691: "the if" -> "whether"

ReviewChanges.js:
- Line 50: sampleValues -> surveyQuestions (to match param extraction in projects.clj)
- Line 100: surveyQuestionUpdated -> updateSurvey (to match param extraction in projects.clj)

project.sql:
- Drop `classification_time` from `projects` table if it is no longer in use.
- Lines 335-346: Delete old function body.

sirmspencer:
projects.clj:

Lines 675-676: No need to bind original-project. Just use the SQL call as the if's test expression.
- I use this binding in the next PR.  Originally I had been checking the difference for the survey questions on the back end but then I noticed the survey questions can be different without any changes (the old survey questtion format is conformed to the new format on the front end).

project.sql:

Drop classification_time from projects table if it is no longer in use.
- I added issue 875

PR #842, opened 2020-10-28
Update message for csv

sirmspencer:


lambdatronic:
projects.clj:
- Line 371: `(set/superset? must-include header-set)`
- Line 373: `must-include` should be a set for `set/superset?` to work
  on line 371. Add `must-include (set must-include)` to the `let`
  bindings after `header-set`. Then drop the `set` call on this line.
- Line 374: Remove extra space after period.
- Line 376: I don't think it is possible to ever return true for this
  conditional branch. If `header-row` contains no commas, then
  `header-set` will have only one entry. Since `must-include` for all
  file types is always more than one entry, the first branch would
  always trigger in this case and throw its error message,
  short-circuiting the `cond`. Thus, this `cond` should simplify to a
  single `if` expression. If you do want to report missing commas in
  the `header-row`, you should make this the first branch in the
  `cond`.

lambdatronic:
projects.clj:
- Lines 361-368: Align columns.
- Line 369: `set headers` -> `(set headers)`

PR #850, opened 2020-10-29
Add new imgaery type for NICFI

sirmspencer:


lambdatronic:
collection.js:
- Line 1080: No showGroup value or toggleShow function.
- Line 1466: `return imagery && imagery.sourceConfig && {`

collectionMenuControls.js:
- Line 227: `htmlFor` must correspond to an id on a form element. Add one to the `select`.
- Line 277: No need to specify `type="button"` on a `<button>` element. This attribute is for `<input>` elements.

CreateProjectWizard.js:
- Line 322: Why did you add `this.steps.imagery` to the return value of `getSteps`?

review-institution.js:
- Line 749: Why is our sourceConfig.type still camelCased as MapBox?
- Line 751: Missing semicolon.

sirmspencer:
Line 1080: No showGroup value or toggleShow function.
-- None are needed here because there is no content to hide and show.

Line 277: No need to specify type="button" on a <button> element. This attribute is for <input> elements.
-- One of the default lint rule requires it. Its more clear than assuming the type (which could be `submit`)

Line 322: Why did you add this.steps.imagery to the return value of getSteps?
-- Imagery has been able to be updated.  It was an oversight that it wasnt already returned with getSteps.  

Line 749: Why is our sourceConfig.type still camelCased as MapBox?
-- Because we already had MapBox imagery in the DB.

PR #851, opened 2020-10-29
Add planet NCFI type for master

sirmspencer:


lambdatronic:
collection.js:
- Line 1391: `return imagery && imagery.sourceConfig && {`

collectionMenuControls.js:
- Line 227: `htmlFor` must correspond to an id on a form element. Add one to the `select`.
- Line 277: No need to specify `type="button"` on a `<button>` element. This attribute is for `<input>` elements.

review-institution.js:
- Line 737: Why is our sourceConfig.type still camelCased as MapBox?
- Line 739: Missing semicolon.

sirmspencer:
Line 277: No need to specify type="button" on a <button> element. This attribute is for <input> elements.
-- One of the default lint rules requires the specificity

Line 737: Why is our sourceConfig.type still camelCased as MapBox?
-- Because we have existing imagery in the DB with type MapBox

PR #852, opened 2020-10-29
Update params for server

### LINKED ISSUE #849: Change :run-server alias to specify both HTTP and HTTPS ports
Short form:
$ clojure -A:run-server -p 8080 -m dev
$ clojure -A:run-server -p 8080 -P 8443 -m prod

Long form:
$ clojure -A:run-server --http-port 8080 --mode dev
$ clojure -A:run-server --http-port 8080 --https-port 8443 --mode prod

Note that we should support using either development or production
mode with or without HTTPS. If -P (--https-port) is specified, then
use SSL/TLS, otherwise don't.

Use clojure.tools.cli/parse-opts to process arguments. See
https://github.com/pyregence/geosync/blob/master/src/geosync/core.clj
for example code.

sirmspencer:
closes #849

### DIFF
--- a/README.org
+++ b/README.org
@@ -94,17 +94,17 @@ To compile and run the web application, navigate to the toplevel
 project directory and run:
 
 #+begin_src sh
-npm run webpack-[dev|prod]
-clojure -A:run-server -m [dev|prod]
+npm run webpack-<dev|prod>
+clojure -A:run-server [-m <dev|prod>] [-p 8080] [-P 8443]
 #+end_src
 
-The website will then be available at http://localhost:8080 unless
-a port is specified.  A http port can be specified with -p and a https
-port can be specified with -P. In dev mode,
-server-side exceptions will be displayed in the browser and Clojure
-source files will be reloaded whenever you refresh the page. These
-features are disabled in prod mode. If the second argument to
-run-server is omitted, it will default to prod mode.
+The website will then be available at http://localhost:8080 unless a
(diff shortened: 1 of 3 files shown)


lambdatronic:
server.clj:
- Lines 39,42: Allow -p and -P to be integers as well, so that
  `start-server!` can be called more easily from the REPL.
```clojure
:parse-fn #(if (int? %) % (Integer/parseInt %))
```
- Line 41: "Port for https (e.g. 8443)"
- Line 46: `#(or (= % "prod") (= % "dev"))` -> `#{"prod" "dev"}`
- Line 49: No need for a dummy `_` binding in map destructuring. These
  are only useful for positional arguments in list destructuring.
- Lines 59-61: Don't tie https to mode=prod. We specifically need to
  be able to run production mode with http only on our Raspberry Pi.
  This also makes it possible to test advanced mode compilation on our
  local machines without https. Also note that the :parse-fn and
  :validation functions ensure that :https-port will be an integer in
  range if specified.
```clojure
https-port (:https-port options)
```
- Line 65: `(when (and https-port has-key?)`
- Line 71: `(if (and https-port (not has-key?))`
- Lines 72-77: Remove mode-specific conditional logic:
```clojure
(println "ERROR:\n"
         " An SSL key is required if an HTTPS port is specified.\n"
         " Create an SSL key for HTTPS or run without the --https-port (-P) option.")
```
  
README.org:
- Update documentation for :run-server alias.

PR #853, opened 2020-10-29
Toggle SIG imagery visibility

sirmspencer:


lambdatronic:
review-institution.js:
- Line 441: Missing semicolon.
- Line 411: Checking for institutionId === 3 feels very brittle. Is
  SIG also institutionId=3 on ceodev.servirglobal.net? Also note that
  this will break every new installation of CEO. A simpler solution
  would be to just allow any institution to make their imagery sources
  publicly available.
- Line 415: institution's -> institutions'

imagery.sql:
- Line 80: Fix comment.

PR #880, opened 2020-11-09
Even more CSV error message updates

sirmspencer:


lambdatronic:
projects.clj:
- Lines 363-369: Don't use stateful operations inside functions run within pure sequence HOFs:
```clojure
:else
(if-let [invalid-headers (seq (remove #(re-matches #"^[a-zA-Z_][a-zA-Z0-9_]*$" %) headers))]
  (init-throw (str "One or more CSV columns are invalid: " (str/join ", " invalid-headers)))
  (do
    (spit ext-file (str/replace-first data header-row (str/join ", " headers)))
    (type-columns headers)))
```
- Line 411: Why did you remove the list of required fields from the Malformed CSV error message?

sirmspencer:
Line 411: Why did you remove the list of required fields from the Malformed CSV error message?
-- The user gets the whole "stack trace".  The malformed message is no longer helpful because the new header messages are more specific.

PR #884, opened 2020-11-10
block scinetific notation for survey answers

sirmspencer:
JS does not fire an on change event for scientific notation, even though a number input allows it. Block this to avoid confusing error messaging.

### DIFF
--- a/src/js/project/SurveyQuestions.js
+++ b/src/js/project/SurveyQuestions.js
@@ -299,6 +299,9 @@ class NewAnswerDesigner extends React.Component {
                         className="value-name"
                         autoComplete="off"
                         value={this.state.newAnswerText}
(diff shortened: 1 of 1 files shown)


lambdatronic:
SurveyQuestions.js:
- Line 303: It's probably safer to simply block any non-numeric
  characters. Otherwise, someone might input a unicode PI if asking a
  question about circles in radians. Or they might try to designate
  units like 30m (meters).

sirmspencer:
Line 303: It's probably safer to simply block any non-numeric
characters. Otherwise, someone might input a unicode PI if asking a
question about circles in radians. Or they might try to designate
units like 30m (meters).
-- I already think blocking e is overkill. If you want to block more can you create an issue?

PR #892, opened 2020-11-11
Update handler stack

sirmspencer:
I noticed that the handler stack for production has ssl-redirect, which means not having a cert with -m prod will still not work.  In looking at the two handler stacks, the only difference is ssl-redirect for prod, and reload for dev.

In your review of the run-server param updates you said
"Don't tie https to mode=prod. We specifically need to
be able to run production mode with http only on our Raspberry Pi.
This also makes it possible to test advanced mode compilation on our
local machines without https."

However, the back end does not have "advanced mode compilation". So we just need to use the production handler stack without ssl-redirect.  we will have to redefine the production handler stack to take a ssl? param.  Most of our middleware is the same in wrap-common handler.  We can also have a reload? param and consolidate into one handler stack function.

Something like (in handler.clj)
```clojure
(defn create-handler-sack [ssl? reload?]
  (as-> authenticated-routing-handler handler
    (if ssl?
      (wrap-ssl-redirect handler)
      handler)
    (wrap-common handler)
    (if reload?
      (wrap-reload handler)
      handler)))
```
and (in server.cljs)
```
handler #(create-handler-sack (and has-key? https-port) (= mode "dev"))
```

This code change appears to be working.  Im not sure what the '# notation was for, so Im not sure that this is a 1 for 1 update.

### DIFF
--- a/src/clj/org/openforis/ceo/handler.clj
+++ b/src/clj/org/openforis/ceo/handler.clj
@@ -152,7 +152,7 @@
           (data-response cause {:status (or status 500)}))))))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-;; Handler Stacks
+;; Handler Stack
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 (defn optional-middleware [handler mw use?]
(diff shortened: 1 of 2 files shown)


lambdatronic:
handler.clj:
- Line 155: Handler Stack
- Line 163: create-handler-sack -> create-handler-stack

server.clj:
- Lines 6,58: create-handler-sack -> create-handler-stack

In response to your question about the #'app syntax, that is the
reader macro for `(var app)`. ring-jetty is built such that if you
pass a handler function as its :handler attribute, then that function
object will be installed in the running server, and off you go.
However, if you redefine that function (e.g., development-app), you
will need to restart the server for your changes to take effect. If
you pass a var that is bound to a handler function as the :handler
attribute to ring-jetty, then redefining the var should allow you to
immediately use the new code changes without restarting the server.

Having said that, it is possible that wrap-reload may also take care
of this via a different route. It would be good to verify this.

PR #895, opened 2020-11-13
Add subtext for user drawn samples

sirmspencer:


lambdatronic:
SampleDesign.js:
- Line 142: lines -> linestrings  (We don't use lines in GIS. They are infinite and defined by exactly two points.)

PR #900, opened 2020-11-17
Update user role process

sirmspencer:


lambdatronic:
users.clj:
- Line 182: `account-id` can be `nil` if `get_user` doesn't return any
  results. `pos?` throws a NullPointerException on nil. To avoid this
  issue and errors where account-id is passed to other `call-sql`
  calls later in the function body, you should add -1 as the default
  return value on line 179:
```clojure
(-> (call-sql "get_user" new-user-email) (first) (:user_id -1))
```
- Line 185: `(nil? email)` is more specific.

PR #909, opened 2020-11-25
Update eslint to use @babel/eslint-plugin

sirmspencer:


rfontanarosa (approved):
It works.

Schermata 2020-12-07 alle 17 43 35

PR #912, opened 2020-11-30
Add OSM as an imagery option

sirmspencer:


rfontanarosa (approved):
It works.

Schermata 2020-12-07 alle 17 16 48

PR #934, opened 2020-12-01
Update boundary check for round numbers

sirmspencer:


lambdatronic:
projects.clj:
- Line 712: These aren't multi-polygons. A single polygon is defined
  as a list of rings. The first ring is called the "exterior ring" and
  describes the outer perimeter of the polygon. All other rings in the
  polygon are called "interior rings" and represent pieces that are
  missing from the polygon shape. This allows a shape to contain
  holes. Your `extract-coords` function only grabs the exterior ring
  from the polygon that is passed to it.
- Line 716: No need for `sort-by`. Your `juxt` function reduces to this:
```clojure
(sort)
```
  However, you shouldn't sort these coords. See below for a correct
  implementation.
- Line 719: Second `geom1` should be `geom2`. Also, you don't need to
  use `get` in this function unless you just like how it reads.
- Line 720: This is a naive comparison between two polygons. Two rings
  may contain the same coordinates, but if they are in a different
  order that is not simply cyclically out of step, then these are
  different polygons. Consider the simple case of a rectangle and a
  bowtie shape. Both can be defined with the same 5 coords (with the
  first and last being the same) but these would be different
  polygons. Here is a robust comparison between exterior rings that
  will correctly match polygon coordinates even if they are reversed,
  cyclically offset, or both reversed and cyclically offset:
 ```clojure
(defn- exterior-ring [coords]
  (map (fn [[a b]] [(tc/val->double a) (tc/val->double b)])
       (first coords)))

(defn- same-ring? [[start1 :as ring1] ring2]
  (or (= ring1 ring2)
      (= ring1 (reverse ring2))
      (= ring1 (take (count ring1) (drop-while #(not= start1 %) (cycle (rest ring2)))))
      (= ring1 (take (count ring1) (drop-while #(not= start1 %) (cycle (reverse (rest ring2))))))))

;; NOTE: This only works for polygons (and only compares their
;; exterior rings). If you need to compare linestrings or points, you
;; will need a different function.
(defn- same-polygon-boundary? [geom1 geom2]
  (and (= "polygon" (str/lower-case (geom1 "type")) (str/lower-case (geom2 "type")))
       (same-ring? (exterior-ring (geom1 "coordinates"))
                   (exterior-ring (geom2 "coordinates")))))
```
- Line 784: same-geom? -> same-polygon-boundary?

PR #937, opened 2020-12-01
Shorten all src/clj/ aliases from org.openforis.ceo.* to just ceo.*

### LINKED ISSUE #806: Shorten all src/clj/ aliases from org.openforis.ceo.* to just collect-earth-online.*
This requires that you match the directory structure as well.

sirmspencer:
closes #806


sirmspencer:
This needs to be done on a clean master

PR #948, opened 2020-12-03
added blog tab to navigation bar

gtondapu:


sirmspencer:
resources/public/img/institution-logos/institution-1.gif
-- This file should not be part of the PR.  It looks like resources/public/img/institution-logos needs to be in gitignore.  Please remove this from your commit and gitignore (or I can do the update, your choice)

src/js/components/PageComponents.js
195, put spaces around your operators `:`, `=`, `?`, etc
195, Since you are comaring directly to a string you define, you dont need toLowerCase. `page === "Blog"`

gtondapu:
Made changes to src/js/components/PageComponents.js
Can you please do the gitignore part?

PR #952, opened 2020-12-04
added a button to refresh basemap list

### LINKED ISSUE #673: Add refresh button next to base map imagery on imagery widgets
On all of the widgets that show the link to create imagery from the institution page, add a "refresh" button next to the drop down.

rfontanarosa:
fix #673

### DIFF
--- a/src/js/widget-layout-editor.js
+++ b/src/js/widget-layout-editor.js
@@ -938,9 +938,9 @@ class WidgetLayoutEditor extends React.PureComponent {
                         id="widgetIndicesSelect"
                         onChange={this.onDataBaseMapSelectChanged}
                     >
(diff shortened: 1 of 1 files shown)


sirmspencer:
src/js/widget-layout-editor.js
60 getInstitutionImagery should be after all life cycle methods (move below componentDidMount)
920-942 I am moving us all to a pattern that works better with eslint.  If you are returning a code block that starts with an html tag use parens.
```
return (
	<div>
	</div>
)
```
938 you can write this shorter `{(this.state.imagery || []).length > 0 && this.state.imagery.map(imagery =>`
I already have a helper function to do that in generalUtils. So the result is even shorter `{safeLength(this.state.imagery) > 0 && this.state.imagery.map(imagery =>`
939 if you want, you can destructure imagery in the function call `.map(({id, title}) => <option key={id} value={id}>{title}</option>)`

Issue #955, opened 2020-12-04
If WMS "Additional WMS Params (as JSON object)" does not have wrapping {}, add them.


sirmspencer:
The two scenarios to test are
1. A blank input box should result in {} which is a valid entry
2. Comma separated fields should add wrapping {} `"var": 12, "num": 8` -> `{"var": 12, "num": 8}`
3. Comma separated fields, missing quotes should add quotes `{var: 12, num: 8}` -> `{"var": 12, "num": 8}`

rfontanarosa:
no problem with the first one, we can use the sanitizer logic.

for the second one it's a tricky thing, we can sanitize it in the same function but we can't use keys without double quotes (the current validator wants them). we can use a complex function to creare a valid json from "var: 12, num: 8" like this:

`{${url.split(",").filter(v => v !== "").map(v => v.split(":")).map(([k, v]) => `"${k.replace(/[ "']/g, "")}": ${v}`).join(",")}}`

sirmspencer:
That sounds good so far.  You can create regex patterns to test against to know how to sanitize the data.

sirmspencer:
Use 
```
sanitizer: json => `{${json
                    .trim()
                    .replace(/[{}]/g, "")
                    .split(",")
                    .filter(u => u !== "")
                    .map(u => {
                        const [k, v] = u.split(":");
                        return v && `"${k.replace(/[ "']/g, "")}": "${v.replace(/[ "']/g, "")}"`;
                    })
                    .join(",")}}"`,
```

PR #957, opened 2020-12-05
added sanitizing logic, added a validation url for GeoServer imagery …

### LINKED ISSUE #682: Check for ? and & in WMS url
These are invalid characters for a WMS (geoserver) url, the query params are supposed to be in the json params input box for a new imagery.  
- [x] If there is just a ? at the end, remove it.  
- [x] If there is stuff after the ? then warn the user to update their imagery.

rfontanarosa:
fix #682

the idea is to have a generic sanitizing logic that could be applied automatically at submit time.

the validation logic is already implemented, I've just added a GeoServer condition.

### DIFF
--- a/src/js/imagery/imageryOptions.js
+++ b/src/js/imagery/imageryOptions.js
@@ -8,7 +8,7 @@ export const imageryOptions = [
             {
                 key: "geoserverUrl",
                 display: "WMS URL",
-                sanitizer: url => url && url.length !== 0 && url.slice(-1) === "?" ? url.slice(0, -1) : url,
+                sanitizer: url => url.endsWith("?") && url.slice(-1) === "?" ? url.slice(0, -1) : url,
             },
             {key: "LAYERS", display: "WMS Layer Name", parent: "geoserverParams"},
             {
--- a/src/js/imagery/imageryOptions.js
+++ b/src/js/imagery/imageryOptions.js
@@ -8,9 +8,7 @@ export const imageryOptions = [
             {
                 key: "geoserverUrl",
                 display: "WMS URL",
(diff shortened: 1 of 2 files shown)


sirmspencer:
11 You wont need more than one sanitizer. The function can do as many checks as it needs to. Change this to `sanitizer:` and dont use an array
557 Instead of creating a new variable, just update 
656 move this above uploadCustomImagery so it is defined before it is called.
657 In this context you should be using forEach.  every is a function that returns true and false.
659 You are mutating sourceConfig.  Create a copy before hand to make this function immutable over all. Use `const sourceConfigCopy = {...sourceConfig};`
666 This is fine for now. In a separate PR, we should move all of these validations into the imageryOptions definition like you did with the sanitizer.

sirmspencer:
11 This can be simplified. `url => url.endsWith("?") ? url.slice(0, -1) : url` url should never be null, are you finding otherwise?
558 Dont use a ternary because you are not returning anything.  Use if statements for side affects.
```
if (o.sanitizer) {
    sourceConfigCopy[o.key] = o.sanitizer(sourceConfigCopy[o.key])
}
```
667 I prefer a period over !. The app does not need to be yelling at the user. "The field \"WMS Url\" should not contain the query string. Please put those values in the field \"Additional WMS Params (as JSON object) \""

rfontanarosa:
fix #955

sirmspencer:
11 You have a unecessary slice `url => url.endsWith("?") ? url.slice(0, -1) : url`
19 can you take this out and make another PR.  Ive been checking the logic more thouroghly and it looks like it needs some work.  I am getting empty {} for most inputs.

PR #958, opened 2020-12-05
Store institution logos as bytea

### LINKED ISSUE #808: Remove resources/public/img/institution-logos/ and store logos in Postgres using base64 encoding
- Look at how we upload shape files for an example to convert the image to base64 (if it isnt already)
- Convert base64 string to bytea using decode https://www.base64decode.net/postgresql-decode
```
insert into table_name (image)
values decode('AcAAFBAO5Az....AQAAAFBCO5gT/AEAABT', 'base64')
```
- Institutions already has a bytea column `logo_data`, use that column
- There is an equivalent encode function when reading the bytea data out of the database https://www.base64encode.net/postgresql-encode
- The safe image component can simplify.  Show the back up image when no base64 data is available.  Show the base64 image with ``
- Create a migration script for all of the existing images. byta('image.jpg') will work for existing images.

sirmspencer:
Note to reviewer.  update-institution and get-institution-details can return an error message without modifying the front end because the response is not handled either way.  With our route security, a user cant get to a page with a missing institution anyway.

closes #808


lambdatronic:
institutions.clj:
- Line 37: No "with" in error message.
- Line 45: `(call-sql "add_institution" name description url)`

users.clj:
- Line 199: select_institution_by_id(institution_id, user_id)

create-institution.js:
- Lines 14,94: newInstitutionsDetails.logo isn't used in this file

review-institution.js:
- Line 140: `logo: ""` -> `institutionAdmin: false`
- Line 163: `logo: ""` -> `institutionAdmin: data.institutionAdmin``
- Lines 171-172: Update these attributes to represent the new values returned by /get-institution-details
- Line 288: Use image/* instead of */* for a more specific media type

member.sql:
- Line 246: Remove comment about institution_return.
- Lines 249,303: table -> TABLE

project.sql:
- Lines 847,876: table -> TABLE

sirmspencer:
review-institution.js:
Line 140: logo: "" -> institutionAdmin: false
Line 163: logo: "" -> `institutionAdmin: data.institutionAdmin``
- Good catch.  Although I dont think newInstitutionDetails.institutionAdmin is used.  It cannot be changed in the UI.

Line 288: Use image/* instead of / for a more specific media type
- This matches the base 64 tag that is passed to the back end.

PR #966, opened 2020-12-08
Justine update data collection wizard (text + images) 

justineb17:


sirmspencer:
resources/public/locale/collection/en.json
15, 19 use double quotes `\"` instead of single, to be consistant.
39, A user has reported that this slide is hard to ready. Is it possible to split into two slides so each image can be less than a screen shot of the whole page? Or maybe try resizing your browser so the screen shot is smaller.

PR #969, opened 2020-12-09
fixed slider input type overlap

### LINKED ISSUE #862: Issues with Planet monthly UI
Interface alignment is incorrect:

image

rfontanarosa:
fix #862

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -10,8 +10,12 @@
     padding: 5px;
     text-align: center;
 }
-.slidecontainer {
+.slide-container {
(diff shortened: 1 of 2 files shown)


sirmspencer:
src/js/imagery/collectionMenuControls.js
705, do we need a div with no class here?  I think you can remove this div tag.

rfontanarosa:
> src/js/imagery/collectionMenuControls.js
> 705, do we need a div with no class here? I think you can remove this div tag.

sorry, I left it only to limitate the changed lines. :)

PR #970, opened 2020-12-10
new validateParams logic

### LINKED ISSUE #965: Move validation into imageryOptions definition
Just like you added the sanitizer functions, move the validation into the definition.

rfontanarosa:
fix #965

### DIFF
--- a/src/js/imagery/imageryOptions.js
+++ b/src/js/imagery/imageryOptions.js
@@ -27,10 +27,10 @@ export const imageryOptions = [
             {key: "LAYERS", display: "WMS Layer Name", parent: "geoserverParams"},
             {
                 key: "geoserverParams",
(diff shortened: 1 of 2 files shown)


sirmspencer:
1/2 Move the && from the end of 1 to the begining of 2.  Its easier to find the operator when its at the begining of the line.
4-16, 157, 187  I dont see any comparisons that rely on more than one param (like dateRangeValidator).  Split these into individual validators and apply at the param level.

sirmspencer:
130 Use outOfRange here `validator: value => value.toString().length !== 4 || outOfRange(value, 2014, new Date().getFullYear()) ?`
165 use outOfRange

sirmspencer:
33, 224, 254, You can make these messages more specific now that they are attached to each param. `Invalid JSON in the \"Additional WMS Params\" field.`

PR #973, opened 2020-12-11
Update geo dash plot route

sirmspencer:


lambdatronic:
geo-dash.js:
- Line 1179: &plotId -> plotId

project.sql:
- Lines 1190,1481: RETURNS table -> RETURNS TABLE

PR #974, opened 2020-12-11
moved MANDATORY validation logic inside validateParams

rfontanarosa:
now almost all the validation logic is inside one function.

### DIFF
--- a/src/js/review-institution.js
+++ b/src/js/review-institution.js
@@ -557,7 +557,10 @@ class NewImagery extends React.Component {
             param.validator && param.validator(imageryParams[param.key])
         );
         const imageryError = imageryOptions[type].validator && imageryOptions[type].validator(imageryParams);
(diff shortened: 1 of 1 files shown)


sirmspencer:
src/js/review-institution.js
560-562, you are already mapping over imageryOptions[type].params. on line 556. Update 557 to `(param.validator && param.validator(imageryParams[param.key])) || ((param.required === false || (imageryParams[param.key] && imageryParams[param.key].length > 0)) && param.display + " is required.")`
573 "You must include a title and attribution."

rfontanarosa:
I've inverted the logic to avoid that the function returns true values.
```
// || ((param.required === false || (imageryParams[param.key] && imageryParams[param.key].length > 0)) || `${param.display} is required.`)
|| ((param.required !== false && (!imageryParams[param.key] || imageryParams[param.key].length === 0)) && `${param.display} is required.`)
```

PR #976, opened 2020-12-12
Update project list routes

sirmspencer:


src/sql/functions/project.sql
```diff
@@ -777,13 +777,10 @@ CREATE OR REPLACE FUNCTION valid_boundary(_boundary geometry)
...
-        WHERE ST_IsValid(_boundary)
-            AND NOT (ST_XMax(_boundary) > 180
-                OR ST_XMin(_boundary) < -180
-                OR ST_YMax(_boundary) > 90
-                OR ST_YMin(_boundary) < -90
-                OR ST_XMax(_boundary) <= ST_XMin(_boundary)
-                OR ST_YMax(_boundary) <= ST_YMin(_boundary))
+        WHERE _boundary IS NOT NULL
+            AND ST_Contains(ST_MakeEnvelope(-180, -90, 180, 90, 4326), _boundary)
+            AND ST_XMax(_boundary) > ST_XMin(_boundary)
```
rfontanarosa, line 782:
```suggestion
            AND ST_XMax(_boundary) > ST_XMin(_boundary)
```

src/sql/functions/project.sql
```diff
@@ -777,13 +777,10 @@ CREATE OR REPLACE FUNCTION valid_boundary(_boundary geometry)
...
+            AND ST_YMax(_boundary) > ST_YMin(_boundary)
```
rfontanarosa, line 783:
```suggestion
            AND ST_YMax(_boundary) > ST_YMin(_boundary)
```
why no more <=?

src/sql/changes/update_2021-01-15_sql_clean.sql
```diff
@@ -0,0 +1 @@
+DROP TYPE project_return;
```
rfontanarosa, line 1: where are all the other changes?

sirmspencer: Functions are automatically dropped and recreated now.  Did I miss any changes that are not just changes to functions?

src/sql/functions/project.sql
sirmspencer: Min and max cant be the same.  ST_IsValid was catching that before so the logic was fine before.

src/sql/functions/project.sql
sirmspencer: Min and max cant be the same.  ST_IsValid was catching that before so the logic was fine before.

sirmspencer: Min and max cant be the same.  ST_IsValid was catching that before so the logic was fine before.

src/sql/changes/update_2021-01-15_sql_clean.sql
sirmspencer: I added `ROP VIEW IF EXISTS project_boundary CASCADE;`

PR #979, opened 2020-12-13
Use simple stats for institution projects

sirmspencer:


rfontanarosa (approved):
tested. it's working.
Schermata 2021-01-27 alle 16 43 06

PR #980, opened 2020-12-13
div buttons

sirmspencer:


src/js/review-institution.js
```diff
@@ -945,42 +945,36 @@ function Project({project, isAdmin}) {
...
+                        className="btn btn-sm btn-outline-yellow btn-block px-3 mr-1"
+                        title="Edit Project"
+                        onClick={() => window.location = `/review-project?projectId=${project.id}`}
+                    >
+                        <UnicodeIcon icon="edit"/>
+                    </span>
+                    <span
+                        className="btn btn-sm btn-outline-red btn-block px-3 mt-0 mr-1"
+                        title="Delete Project"
+                        onClick={() => this.props.deleteProject(project.id)}
```
rfontanarosa, line 959: this.props is not defined. use deleteProject.

rfontanarosa (changes requested):
- in create-institution.js line 21 there is a call of a not definied method: this.getInstitutions();
- in review-institution.js line 922 deleteProject is not destructured from props 
- in review-institution.js line 959 change no longer valid this.props.deleteProject with deleteProject

PR #982, opened 2020-12-14
"additional WMS params" sanitizer

### LINKED ISSUE #955: If WMS "Additional WMS Params (as JSON object)" does not have wrapping {}, add them.

rfontanarosa:
fix #955


sirmspencer:
31, remove comment that no longer applies.

Issue #827, opened 2020-10-28
Update the links on the survey question tree that used to go to the survey rule


sirmspencer:
In src/js/project/SurveyCardList.js `function SurveyQuestionTree`
- [x] Remove link that no longer works.  The rules are on a separate component so the scroll to links dont work

- [x] Add rule information to tool tip.  This can be done by setting the attribute `title="This is a tool tip"` for a basic slow tool tip.  Or follow the example here https://www.w3schools.com/css/css_tooltip.asp for a nice tootip.

PR #983, opened 2020-12-16
Githika update surveylinks

gtondapu:
Added the tooltips to survey links(no styling).


gtondapu:
https://github.com/openforis/collect-earth-online/issues/827

sirmspencer:
src/css/custom.css
This looks good so far.  Can you scroll down on the page I linked to the top arrow section and use that?
657 you dont need dotted block underline because you have color.

55 add space before function
55-67 Why are you not using r.ruleType to decide which message to show?
55-67 I dont think your messages need to start with `"Rule: " + r.ruleType + "` here because the main text that you are moving your mouse to already include this information.
145 Our lint rules require a `,` for the last item in a list.
191 move the color style to .surveyrule_tooltip
191 where are your classes being applied for the tool tips?
191 When would you ever return ""?  You are only adding tool tips where there are already rules so a message "No rules." does not makes sense.  If getRulesById() doesnt have anything to return, add the error message should be returned from there instead.

Can you post a screen shot of a working tool tip?

gtondapu:
I am filtering with rule id which gives the corresponding rule, we can also add ruleType but I thought it was not required.
I had the classes in css as placeholders for tooltip styling but I ended up adding a basic tooltip, so removed those now.
Edited the rule messages to not "Rule: " + r.ruleType + "
Yes I agree, "No rules" message does not make sense.

Issue #985, opened 2020-12-16
Fix formatting for survey rule designer


sirmspencer:
![image](https://user-images.githubusercontent.com/33734037/102375071-ea07c280-3f76-11eb-9fc0-0051a568c1bc.png)

PR #993, opened 2020-12-16
Githika ruleformatting

gtondapu:
https://github.com/openforis/collect-earth-online/issues/985


sirmspencer:
src/css/custom.css
654-681 I think only surveyrule_select is used in this PR. surveyrule_select only includes `overflow: scroll`. I think you can just use the bootstrap class `overflow-auto`

src/js/project/SurveyRules.js
392, 438, 456 surveyrule_select only includes `overflow: scroll`. I think you can just use the bootstrap class `overflow-auto`

Issue #803, opened 2020-10-28
Add a minimal initial SQL dataset for development


sirmspencer:
Add to src/sql/default_data/default_data.sql
- Guest user 
user_uid = -1
email = guest 
password = null
admin = false

Add to new file src/sql/default_data/dev_data.sql
- `-- NAMESPACE: dev-data` at the top
- Site Admin User
email = admin
password = admin
admin = true
- Normal user
email = user
password = user
admin = false
- 1 public imagery
      {"id": 1,
       "institution": 1,
       "visibility": "public",
       "title": "Open Street Maps",
       "attribution": "Open Street Maps",
       "extent": null,
       "sourceConfig": {"type": "OSM"},
- 1 Institution (use 'user')
     - 1 Project
       - 1 GeoDash Configuration
       - 3 Plots
         - 10 Samples Per Plot

PR #994, opened 2020-12-16
dev-data command to import basic data

rfontanarosa:


sirmspencer:
src/sql/default_data/default_data.sql
13 collect.earth is our domain.  Since it cannot be accessed by anyone who uses this repo, do not use the domain. No domain is ok, use `guest`
13 we dont want anyone to log in as guest. Make the password column '', not the password. (ie, dont use crypt here)

src/sql/dev_data/dev_data.sql
7 collect.earth is our domain.  Since it cannot be accessed by anyone who uses this repo, do not use the domain. No domain is ok, use `admin`
src/sql/dev_data/dev_data.sql
13 collect.earth is our domain.  Since it cannot be accessed by anyone who uses this repo, do not use the domain. No domain is ok, use `user`
14, 20, 27, 33, 43, 49, 107, 113 Anytime we add items with uid > 0 to a table with a serial primary key, we need to update the index. `SELECT setval(pg_get_serial_sequence('"table_name"', '"primary_column"', [max_id + 1]))`
19 use just 'institution-3.png' the front end code adds img/institution-logos/
36-42 I think it will be easier to read if you add a return for every item instead of just for survey questions.
36-42 You dont need to add values if the value is null or is the same as its default (the ts_ columns have defaults)

sirmspencer:
src/sql/dev_data/dev_data.sql
setval(pg_get_serial_sequence, did you test these?  I wrote a function a while back for this on another project and the table and column names needed "" like `SELECT setval(pg_get_serial_sequence('"users"', '"user_uid"'), (SELECT MAX(user_uid) FROM users) + 1);`  You can test this by adding an item to a table through the UI (ie create a new institution).  If the index is off you will get a unique id error. 
9, this is uneeded, you have once for the users table in line 9. 

57/79 a sample_resolution on -1 makes no sense.  You dont need a sample_resolution with a 'center' sample distribution.
57/79/111-155 you number of samples does not match a 'center' sample distribution.  More than one sample per plot is good, update your sample distribution to match what you created.
61/83 remove ts_plot_size which has a default, and is not used anywhere in our current code.

sirmspencer:
src/sql/dev_data/dev_data.sql
setval(pg_get_serial_sequence, did you test these?  I wrote a function a while back for this on another project and the table and column names needed "" like `SELECT setval(pg_get_serial_sequence('"users"', '"user_uid"'), (SELECT MAX(user_uid) FROM users) + 1);`  You can test this by adding an item to a table through the UI (ie create a new institution).  If the index is off you will get a unique id error. 
4-15, I just saw this, this doesnt need to be two insert statements.  You can have multiple rows of values.

57/79 a sample_resolution on -1 makes no sense.  You dont need a sample_resolution with a 'center' sample distribution.
57/79/111-155 you number of samples does not match a 'center' sample distribution.  More than one sample per plot is good, update your sample distribution to match what you created.
61/83 remove ts_plot_size which has a default, and is not used anywhere in our current code.

rfontanarosa:
> SELECT setval(pg_get_serial_sequence('"users"', '"user_uid"')

yes, I tried it. I discovered that it works until table/field names are lowercase.

**For compatibility with the handling of ordinary SQL names, the string will be converted to lower case unless it contains double quotes around the sequence name.**

postgresql.org/docs/9.2/functions-sequence.html

sirmspencer:
src/sql/dev_data/dev_data.sql
10 `+ 1` MAX(user_uid) will now be 2 and you just need to add the 1 to it.
78, samples_per_plot is 1, when it should be higher. "random" looks good for the sample_distribution.

rfontanarosa:
fix #803

Issue #647, opened 2020-08-05
Show plot level message for planet daily instead of hiding the imagery from the selection list.


sirmspencer:
Look at secure watch as an example.  At the plot level the map shows an image with instructions to go to a plot before the imagery will be available.

Also there are several place where the planet imagery is filtered as an option. Remove these so that the instructions show instead of having some other imagery selected.

PR #1000, opened 2020-12-22
created layer name GOTOPLOT

rfontanarosa:


sirmspencer:
src/js/collection.js
223-229 I think we can make this more generic by just adding the layer when a map is created.  In mercator.js, in mercator.createMap, line 648, add the goToPlot layer when the map is created. Use visible: false and zIndex: 100.

Use mercator.setLayerVisibilityByLayerId or mercator.setVisibleLayer to hide / show the layer when necessary. 

the layer "id" can be a string.  Make the layer id "goToPlot" insted of -1. -1 in this app means something more like not found, invalid selection, etc.  Its easier to understand what you are doing if the layers id has something meaningful.

JS uses camel case so use goToPlot instead of GOTOPLOT.

268 `currentImagery.sourceConfig.type === "PlanetDaily" || currentImagery.sourceConfig.type === "SecureWatch")` can be simplified to `["PlanetDaily", "SecureWatch"].includes(currentImagery.sourceConfig.type)`

src/js/utils/mercator.js
463-469 I dont think you need a source type here.  You can just add the goToPlot layer when the map is created and hide / show it.
464 Is there a difference between the TileWMS layer and the XYZ layer (see layer 471)? The XYZ layer is one line verses a couple.

sirmspencer:
src/js/utils/mercator.js
652-656 I think you only need url here.  If you need more than url here, use XYZ layer, it only needs url.

rfontanarosa:
fix #647

Issue #856, opened 2020-10-29
example .csv and .shp files


kdyson:
In parallel with #855 , consider providing example .csv and .shp files (known good) that users can use as templates.

Also useful for testing.

sirmspencer:
Can you please create the files that you wish to be used as and example?

sirmspencer:
[example-external-files.zip](https://github.com/openforis/collect-earth-online/files/5703993/example-external-files.zip)

sirmspencer:
- [ ] Remove the old test files, they have errors in them (resources/test_data)
- [ ] Unzip the attached file and add them to resources/public/test_data
- [ ] in /src/clj/org/openforis/ceo/js/project/PlotDesign and /src/clj/org/openforis/ceo/js/project/SampleDesign add a link under each file selector that says "Download example plot shp file".  Only show this link when the file selector is visible.

gtondapu:
I have made the changes. After I did some testing I found the following results. Is everything good with Plot shape file?

uploading plots csv and samples csv - works
uploading plots csv and samples shp - works
uploading plots shp and samples shp - would not work
uploading plots shp and samples csv- would not work

PR #1005, opened 2021-01-03
Githika csvshp files

gtondapu:
https://github.com/openforis/collect-earth-online/issues/856


sirmspencer:
resources/public/test_data/plot-shp.zip
resources/public/test_data/plots-example.csv
resources/public/test_data/sample-shp.zip
resources/public/test_data/samples-example.csv
- Lets be consistant with our names.  plot-shape-example.shp, plot-csv-example.csv, etc

src/js/project/PlotDesign.js
- 226 You should not be nesting a div inside a label.  It looks like you are using the div that wraps the link to jump out of the natural order of div heiarchy. Move the link tag outside the label and use higher level divs to create order.  It looks like the link needs to go outside the div on 200.  This div creates the row for the file name.  Then the link and the div on 200 need a wraping div to create a column of the div and link.
- 226 This line is long
- 226 Use space around all operators including ternary.
<a href={fileType === "csv" ? "test_data/plots-example.csv" : "test_data/plot-shp.zip"}>
    Download example plot {fileType === "csv" ? "csv" : "shape"} file
</a>

src/js/project/SampleDesign.js
Same comments as above

- 192 revert change for tool tips for this branch.

PR #1008, opened 2021-01-04
now it updates only the widgets that have changed their layout

### LINKED ISSUE #659: Updating widgets creates too many calls to the server
Right now every update creates an update call for every widget, not just the one that is updated. And actually this runs twice.  The update is also called on page load.  Lets start by only updating one widget when once when it changes.

rfontanarosa:
fix #659

### DIFF
--- a/src/js/widget-layout-editor.js
+++ b/src/js/widget-layout-editor.js
@@ -164,7 +164,7 @@ class WidgetLayoutEditor extends React.PureComponent {
         });
         this.setState({widgets: widgets});
         if (changed) {
-            this.updateServerWidgets();
+            this.updateAllServerWidgets();
         }
     };
 
@@ -188,7 +188,13 @@ class WidgetLayoutEditor extends React.PureComponent {
             });
     };
 
-    updateServerWidgets = () => {
+    updateServerWidgets = widgets => {
+        widgets.forEach(widget => {
+            this.serveItUp(`${this.state.theURI}/update-widget?widgetId=${widget.id}`, widget);
+        });
+    };
+
+    updateAllServerWidgets = () => {
         this.state.widgets.forEach(widget => {
             this.serveItUp(`${this.state.theURI}/update-widget?widgetId=${widget.id}`, widget);
         });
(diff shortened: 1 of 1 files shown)


sirmspencer:
191-201, it is confusing here to have two functions that do the same thing.  It would be easier to follow to just pass this.state.widgets to updateServerWidgets if that is what is being updated.
1736 it is customary to use destructuring here. `const {widgets} = this.state`
1764 You dont need to create a array just to pass it to updateServerWidgets. Then you are looping over the object twice.
1765 this line is causing mutation. I dont think 1763 is creating new copies of each widget object, just a new array with references to the same widget object.  This means that you are mutating the state directly here.  This logic wont cause mutation.
`const newWidgets = widgets.map((w, i) => ({...w, layout: layout[i]}))`
1769 this does not need to be a call back if you are not using this.state in updateServerWidgets.  
1769 The only reason widgetsToUpdate has correct values for updateServerWidgets is because widgetsToUpdate is array of objects that are the same as widgets.  You then mutate individual widget objects in widgets (that are the same object).  This code can look un-intentional.

Here is one option that does not mutate.  Its not ideal to also have side affects in the map but it works.
```
const {widgets} = this.state.widgets;
const newWidgets = widgets.map((w, i) => {
    if (_.isEqual(w.layout, layout[i])) {
        return w;
    } else {
        const newWidget = {...w, layout: layout[i]};
        this.serveItUp(`${this.state.theURI}/update-widget?widgetId=${newWidget.id}`, newWidget);
        return newWidget;
    }
});
this.setState({
    widgets: newWidgets,
    layout: layout,
});
```

You could also be more simple here and then add code for componentDidUpdate.  The code here would be
```
onLayoutChange = layout => {
    if (this.state.haveWidgets) {
        this.setState({
            widgets: this.state.widgets.map((w, i) => ({...w, layout: layout[i]})),
            layout: layout,
        });
    } else {
        this.setState({layout: layout});
    }
};
```

And then
```
componentDidUpdate(prevProps, prevState) {
    if (prevState.widgets !== this.state.widgets && prevState.widgets.length !== this.state.widgets.length) {
        this.state.widgets.forEach((w, i) => {
            if (!_.isEqual(w.layout, prevState.widgets[i])) {
                this.serveItUp(`${this.state.theURI}/update-widget?widgetId=${w.id}`, w);
            }
        });
    }
}
```

The benifit to the second option is that you can also remove 
if (changed) {
    this.updateAllServerWidgets();
}

src/js/widget-layout-editor.js
```diff
@@ -164,7 +164,7 @@ class WidgetLayoutEditor extends React.PureComponent {
         });
         this.setState({widgets: widgets});
         if (changed) {
-            this.updateAllServerWidgets();
+            this.updateServerWidgets();
```
sirmspencer: Keep as updateAllServerWidgets.  The function you left is updateAllServerWidgets.

PR #1009, opened 2021-01-05
added plot id information on geodash navigation bar

### LINKED ISSUE #641: Include the plot number somewhere on the page with the two graphs.
Clicking 'save' on a plot pops up the Geodash graphs again without it being super clear that it's a new plot. If possible, include the plot number somewhere on the page with the two graphs.

rfontanarosa:
fix #641

### DIFF
--- a/src/js/components/PageComponents.js
+++ b/src/js/components/PageComponents.js
@@ -238,7 +238,7 @@ export class GeoDashNavigationBar extends React.Component {
     });
 
     render() {
-        const {userName, page} = this.props;
+        const {userName, page, plotId} = this.props;
         const uri = window.location.pathname;
 
         return (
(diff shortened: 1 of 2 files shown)


src/js/components/PageComponents.js
```diff
@@ -263,13 +263,18 @@ export class GeoDashNavigationBar extends React.Component {
...
-                            <ul id="login-info" className="navbar-nav mr-auto">
+                        <div className="collapse navbar-collapse justify-content-between" id="navbarSupportedContent">
+                            <ul className="navbar-nav">
                                 <li className="nav-item my-auto">
                                     <h1>GEO-DASH</h1>
                                 </li>
                             </ul>
-                            <ul className="navbar-nav mr-0">
+                            <ul className="navbar-nav" style={{flex: 1, justifyContent: "flex-end"}}>
+                                {uri !== "/widget-layout-editor" &&
```
sirmspencer, line 273: Your JSX below is the inverse logic.  Two options:
1. check for `plotId > 0` 
2. Use a Ternary so you have one conditional with two outcomes together.

src/js/components/PageComponents.js
```diff
@@ -263,13 +263,18 @@ export class GeoDashNavigationBar extends React.Component {
                         >
                             <span className="navbar-toggler-icon"></span>
                         </button>
-                        <div className="collapse navbar-collapse" id="navbarSupportedContent">
-                            <ul id="login-info" className="navbar-nav mr-auto">
+                        <div className="collapse navbar-collapse justify-content-between" id="navbarSupportedContent">
+                            <ul className="navbar-nav">
```
sirmspencer, line 267: Why are all of these items ul with a single li?  Can you can clean this up.

src/js/geo-dash.js
```diff
@@ -1729,6 +1730,7 @@ export function pageInit(args) {
     ReactDOM.render(
         <GeoDashNavigationBar
             userName={args.userName || "guest"}
+            plotId={args.plotId || "0"}
```
sirmspencer, line 1733: - For the most part, CEO uses -1 to mean not a valid selection / value instead of 0.  Here 0 would not be a valid plotid if it was generated by us, but it could be from external data if the user was using a 0 based index.
-   plotId is an integer.
`plotId={parseInt(args.plotId) || -1}`

Issue #1007, opened 2021-01-04
Add icon for questions with rules for hover text


sirmspencer:
On the collection page, in the survey questions, add an icon if there is a rule for that question.  Then use your updated tool tip format to show a nicer looking tool tip.  

The icon should use the new SvgIcon component (see SvgIcon.js).  Search for a free SVG icon on Google that will make sense and add it to SvgIcon.js.  One thing to note here is that if you find and SVG online there is usually extra tags and attributes.  Look at the existing example and try and get the svg information as clean as possible.

PR #1011, opened 2021-01-10
Githika surveyrule icon

gtondapu:
Added icon for survey questions in collection page when a rule exists for a question.
Added format for the tooltip of collection page. Modified tooltip style.
Fixed a quote (') in surveycardlist.js


gtondapu:
https://github.com/openforis/collect-earth-online/issues/1007

src/js/components/SurveyCollection.js
```diff
@@ -350,11 +350,22 @@ class SurveyQuestionTree extends React.Component {
                     >
                         {this.state.showAnswers ? <span>-</span> : <span>+</span>}
                     </button>
+
+                    {this.props.getRulesById(this.props.surveyNode.id).length > 0
```
sirmspencer, line 354: You dont need a ternary here.  Use `this.props.getRulesById(this.props.surveyNode.id).length > 0 &&`

gtondapu: Done made the change and added icon!

sirmspencer:
- You dont need the <g> tags in the SVG. g is for group, if there is not attributes in g it is unecessary.
- Some of the longer set of rules dont display well.  Can you 1) create a list with - or bullets for questions with multiple rules and 2) use a different width for the collection page and the survey design page? It looks like the survey collection panel is narrower than the rule designer.  You may want to pull out other items like left so each location can be styled a little differently.

sirmspencer:
image

PR #1014, opened 2021-01-13
fixes geo dash infinite loop

### LINKED ISSUE #962: addTileServer creates an infinite loop trying to reload the tile
The error function is continuously called if earth engine doesnt return a value.

rfontanarosa:
fix #962

### DIFF
--- a/src/js/geo-dash.js
+++ b/src/js/geo-dash.js
@@ -1077,7 +1077,7 @@ class MapWidget extends React.Component {
                 try {
                     window.setTimeout(function() {
                         error.tile.attempt = error.tile.attempt ? error.tile.attempt + 1 : 1;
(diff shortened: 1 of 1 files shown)


src/js/geo-dash.js
```diff
@@ -1076,9 +1076,9 @@ class MapWidget extends React.Component {
             source.on("tileloaderror", function(error) {
                 try {
                     window.setTimeout(function() {
-                        console.log("trying to reload the tile: " );
-                        console.log(error.tile);
-                        error.tile.load();
+                        error.tile.attempt = error.tile.attempt ? error.tile.attempt + 1 : 1;
+                        // console.log(`Trying to reload Tile ${error.tile.tileCoord.join("-")} (${error.tile.attempt}): `, error.tile);
```
sirmspencer, line 1080: Un comment or remove this line

PR #1015, opened 2021-01-15
Update NICFI layers

sirmspencer:


rfontanarosa (approved):

lambdatronic:
collectionMenuControls.js:
- Line 236: What happens when the user selects "newest"? It looks like
  this is set as the value of `selectedTime` in the `onChange`
  handler, but I don't see where that gets translated back to
  `last(nicfiLayers)`.

Issue #581, opened 2020-07-15
Add loading modals to more places in the site


sirmspencer:
Create project takes a while, and we use a modal to show that the action is in progress.  Make this modal a common component and add it to other locations where stuff is loading.

sirmspencer:
- [ ] institution page initial loading
- [ ] institution page adding or updating a member

sirmspencer:
Use project-admin.js as an example.  Line 118 is the function that wraps a function so that a message will be displayed. Line 141 is where the LoadingModal component is rendered.  Line 56 is the additional state. In ManageProject.js line 136 you can see and example of how to use processModal.

For the home page initial loading, you wont need to create the processModal function because the modal is only needed once in compoenentDidMount.  Just use the same processes as is used internally to processModal.

PR #1021, opened 2021-01-25
Githika modals

gtondapu:
https://github.com/openforis/collect-earth-online/issues/581


sirmspencer:
src/js/home.js
24-39, can we go back to using Promise.all?  The reason we use promise.all is so that the user only gets 1 error message if the process fails instead of up to 3.  .finally should work on Promise.all the same way.
```
        Promise.all([this.getImagery(), this.getInstitutions(), this.getProjects()])
            .catch(response => {
                console.log(response);
                alert("Error retrieving the collection data. See console for details.");
            })
            .finally(() => this.setState({modalMessage: null}));
```
112 remove blank line

src/js/review-institution.js
1076, I beleive we are only updating 1 user at a time "Updating user"
1146 This is unused

PR #1027, opened 2021-02-02
collection_start value is now saved when a plot is flagged

### LINKED ISSUE #665: Add collection start to flag-plots route

rfontanarosa:
fix #665


sirmspencer:
src/clj/org/openforis/ceo/db/plots.clj
148, drop the nil here.  Karen said we will never need the confidence for a flagged plot.
147-149 in clojure, when we can, we column align similar statements.  this includes the right side of the let assignments.
```
(defn flag-plot [{:keys [params]}]
  (let [plot-id          (tc/val->int (:plotId params))
        user-id          (:userId params -1)
        collection-start (tc/val->long (:collectionStart params))]
    (call-sql "flag_plot" plot-id user-id nil (Timestamp. collection-start))
    (unlock-plots user-id)
    (data-response "")))
```


src/sql/changes/update_2021-01-01_flag_collection_start.sql
Drop this file.  `clojure -A:build-db only-functions` will drop and recreate all functions.  I run this every time that I re-deploy the server.

src/sql/functions/project.sql
1141, 1154, 1156 remove _confidence
1162, set confidence to NULL on conflict

PR #1028, opened 2021-02-03
Clean up survey rule designer

sirmspencer:
Errors fixed
- Missing or incomplete validations
    - Example 1 many rules did not check to see if the selected question was valid
    - Example 2 most of the comparisons for lists were insufficient to determine equivalency (i.e. [1].every(e => [1, 2, 3].includes(e)) is true but the two arrays are not equivalent)
- No messages to users.  Errors failed silently
- Silently mutating existing rules without warning the user
- Invalid handling of invalid ui events
    - All of the form components did not have "value" set.  So if a UI change was invalid, the state did not update, but the UI was not reset to match the current (unchanged) state.
- State is not correctly reset when a change dictates.
    - Example 1, when changing from rule type where the different rules share this.state.questionIds
    - When changing a question in incompatible answers, question 2 could be eliminated from the dropdown but q2 and a2 were not reset. When selecting a new question the array of answers was updated without ensuring that the selected answer was a subset.
- Storing easily derived variables in state instead of calculating on submit
- Different types shared single state variables.
- updateOptions uses annoying nested logic without actually eliminating repeat code.

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -1,7 +1,7 @@
 /***********/
 /* General */
 /***********/
-#srd td {
+.srd td {
     border-collapse: collapse;
     border: 2px solid rgb(200, 200, 200);
     letter-spacing: 1px;
--- a/src/js/project/ManageProject.js
+++ b/src/js/project/ManageProject.js
@@ -21,9 +21,11 @@ export default class ManageProject extends React.Component {
     /// API Calls
 
     getProjectDetails = () =>
-        Promise.all([this.getProjectById(this.context.projectId),
-                     this.getProjectImagery(this.context.projectId),
-                     this.getProjectPlots(this.context.projectId)])
+        Promise.all([
+            this.getProjectById(this.context.projectId),
+            this.getProjectImagery(this.context.projectId),
+            this.getProjectPlots(this.context.projectId),
+        ])
             .catch(response => {
                 console.log(response);
                 alert("Error retrieving the project info. See console for details.");
--- a/src/js/project/SurveyRules.js
+++ b/src/js/project/SurveyRules.js
@@ -23,7 +23,9 @@ export class SurveyRuleDesign extends React.Component {
                     inDesignMode
                 />
                 <table id="ruleDesigner">
-                    <label className="font-weight-bold mt-2">New Rule:</label>
+                    <caption style={{color: "black", captionSide: "top", fontWeight: "bold"}}>
+                        New Rule:
+                    </caption>
                     <tbody className="srd">
                         <tr>
                             <td>
@@ -95,8 +97,7 @@ export class SurveyRulesList extends React.Component {
 
         "numeric-range": ({min, max, questionsText}) => (
             <>
-                <td>Min: {min}</td>
-                <td>Max: {max}</td>
+                <td>Min: {min}, Max: {max}</td>
                 <td>Questions: {questionsText.toString()}</td>
             </>
         ),
@@ -612,7 +613,7 @@ export class IncompatibleAnswers extends React.Component {
             ));
         const errorMessages = [
             conflictingRule
-                && "A incompatible answers rule already exists for these question / answer pairs. (Rule "
+                && "An incompatible answers rule already exists for these question / answer pairs. (Rule "
                 + conflictingRule.id
                 + ")",
             (questionId1 === questionId2) && "You must select two different questions.",
--- a/src/js/project/SurveyRules.js
+++ b/src/js/project/SurveyRules.js
@@ -1,6 +1,9 @@
 import React from "react";
+import {isNumber, arraysSameElements} from "../utils/generalUtils";
 
 import {ProjectContext} from "./constants";
+
+const getNextId = (array) => array.reduce((maxId, obj) => Math.max(maxId, obj.id), 0) + 1;
 
 export class SurveyRuleDesign extends React.Component {
     constructor(props) {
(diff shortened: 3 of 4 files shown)


lambdatronic:
generalUtils.js:
- Line 121: Rename `arraysSameElements` to `sameContents`.

SurveyRules.js:
- Line 2: Wrong `import` grouping. Local JS libs should be separated from NPM libs with a blank line.
- Lines 2,375,473,474: Rename `arraysSameElements` to `sameContents`.
- Line 16: Remove extra blank line.
- Line 36: this.selectedRuleType -> this.state.selectedRuleType
- Lines 83-166: Eliminate boilerplate in `render` with refactoring. This is 30 lines shorter and is good FP and DOP:
  ```js
  ruleTypeLabel = {
      "text-match": "Text Regex Match",
      "numeric-range": "Numeric Range",
      "sum-of-answers": "Sum of Answers",
      "matching-sums": "Matching Sums",
      "incompatible-answers": "Incompatible Answers",
  };

  ruleSpecificColumns = {
      "text-match": ({regex, questionsText}) => (<td>Regex: {regex}</td>
                                                 <td colSpan="2">Questions: {questionsText.toString()}</td>),

      "numeric-range": ({min, max, questionsText}) => (<td>Min: {min}</td>
                                                       <td>Max: {max}</td>
                                                       <td colSpan="2">Questions: {questionsText.toString()}</td>),

      "sum-of-answers": ({validSum, questionsText}) => (<td>Valid Sum: {validSum}</td>
                                                        <td colSpan="2">Questions: {questionsText.toString()}</td>),

      "matching-sums": ({questionSetText1, questionSetText2}) => (<td>Questions Set 1: {questionSetText1.toString()}</td>
                                                                  <td colSpan="2">Questions Set 2: {questionSetText2.toString()}</td>),

      "incompatible-answers": ({questionText1, answerText1, questionText2, answerText2}) => (<td>Question 1: {questionText1}, Answer 1: {answerText1}</td>
                                                                                             <td colSpan="2">Question 2: {questionText2}, Answer 2: {answerText2}</td>),
  };

  renderRuleRow = (rule, uid) => {
      const {id, ruleType} = rule;
      return (
          <tr id={"rule" + id} key={uid}>
              {this.props.inDesignMode &&
               <td>{this.removeButton(id)}</td>
              }
              <td>{"Rule " + id}</td>
              <td>Type: this.ruleTypeLabel[ruleType]</td>
              {this.ruleSpecificColumns[ruleType].call(null, rule)}
          </tr>
      );
  };

  render() {
      const {surveyRules, children} = this.props;
      return (
          <div id="survey-rule-design">
              <span className="font-weight-bold">Rules:</span>
              <table id="srd">
                  <tbody>
                      {(surveyRules || []).length > 0
                      ?
                          surveyRules.map(this.renderRuleRow)
                      :
                          <tr>
                              <td colSpan="6">No rules set for this survey</td>
                          </tr>
                      }
                      {children}
                  </tbody>
              </table>
          </div>
      );
  }
  ```
- NOTE: The `inDesignMode` optional column exposes a major flaw in the
  table-based layout used here since `children` and the
  `surveyRules.length === 0` rows both use `colSpan="6"`. This table
  code needs to be replaced with CSS flex code (but the refactored
  structure should be fine as is).
- Line 180: Destructure `this.context` and `this.state` to shorten this function's body:
  ```js
  const {surveyRules, surveyQuestions, setProjectDetails} = this.context;
  const {questionId, regex} = this.state;
  ```
- Line 183: Lowercase "text regex match" to match all the other rule components.
- Line 183: Rule # -> Rule (since we don't use a # in `SurveyRulesList.render`)
- Line 200: Missing semicolon.
- Line 254: colSpan="5" !== colSpan="6"
- Line 271: Destructure `this.context` and `this.state` to shorten this function's body:
  ```js
  const {surveyRules, surveyQuestions, setProjectDetails} = this.context;
  const {questionId, min, max} = this.state;
  ```
- Line 274: Your `conflictingRule` test isn't that specific. The error message should be: "A numeric range rule already exists for this question."
- Line 292: Missing semicolon.
- Line 305: All the other rows in this nested table only have 2 columns. There shouldn't be a `colSpan` attribute here.
- Line 374: Destructure `this.context` and `this.state` to shorten this function's body:
  ```js
  const {surveyRules, surveyQuestions, setProjectDetails} = this.context;
  const {questionIds, validSum} = this.state;
  ```
- Line 396: Missing semicolon.
- Line 413: All the other rows in this nested table only have 2 columns. There shouldn't be a `colSpan` attribute here.
- Line 472: Destructure `this.context` and `this.state` to shorten this function's body:
  ```js
  const {surveyRules, surveyQuestions, setProjectDetails} = this.context;
  const {questionSetIds1, questionSetIds2} = this.state;
  ```
- Line 478: "Matching sums rule requires that both selections contain two or more questions."
- Lines 479,450: Missing periods at the end of the error messages.
- Line 502: Missing semicolon.
- Line 519: All the other rows in this nested table only have 2 columns. There shouldn't be a `colSpan` attribute here.
- Line 596: Destructure `this.context` and `this.state` to shorten this function's body:
  ```js
  const {surveyRules, surveyQuestions, setProjectDetails} = this.context;
  const {questionId1, answerId1, questionId2, answerId2} = this.state;
  ```
- Lines 634,640: Missing semicolon.
- Lines 651,655: All the other rows in this nested table only have 2 columns. There shouldn't be a `colSpan` attribute here.
- Lines 254,358,456,572,738: Missing period at end of label text.

sirmspencer:
NOTE: The inDesignMode optional column exposes a major flaw in the
table-based layout used here since children and the
surveyRules.length === 0 rows both use colSpan="6". This table
code needs to be replaced with CSS flex code (but the refactored
structure should be fine as is).
-- I un nested the two tables so the selection doesnt need to specify colspan and try to stay the same size as the rule list.  It was very temping to keep going but I refrained for the sake of review.

lambdatronic:
SurveyRules.js:
- Line 26: `table` can only contain `thead` and `tbody` sections. Move the `label` tag above the `table`.
- Lines 98-99: Combine min and max properties into a single column to ensure all rows always
               have the same number of columns regardless of ruleType:
  ```html
  <td>Min: {min}, Max: {max}</td>
  ```
- Line 615: A incompatible -> An incompatible

PR #1042, opened 2021-02-06
is now possible add a reason when a plot is flagged

### LINKED ISSUE #76: Add reason for flagging a plot bad
Currently we have the option for users to flag a plot as bad without any justification. It will be good to add a dialog for users to provide an explanation as to why the plot was flagged as bad. These can be simple such as "Cloud cover" or "Poor resolution". This will allow administrators to track bad plots in more detail.

rfontanarosa:
fix #76


sirmspencer:
src/clj/org/openforis/ceo/db/plots.clj
145 align (:fl with the ( above

src/sql/changes/update-2021-02-02_flagging-reason.sql
3-7 There should be no plot_return or plot_collection_return to alter

src/sql/functions/project.sql
1152 use MAX(flagged_reason)
1158 dont group by flagged_reason
-- This code is odd, I know.  When I wrote it a couple years ago we were planning on allowing multiple users collect the same plot.  We are still planning on doing so, so this code reflects that future plan.

PR #1043, opened 2021-02-10
added regex validation for url in xyz imagery type

### LINKED ISSUE #1030: Add validator for new XYZ source

rfontanarosa:
fix #1030

### DIFF
--- a/src/js/imagery/imageryOptions.js
+++ b/src/js/imagery/imageryOptions.js
@@ -66,7 +66,7 @@ export const imageryOptions = [
         params: [{
             key: "url",
             display: "XYZ URL",
-            validator: value => !(/https?:\/\/.*{-?[xyz]}\/{-?[xyz]}\/{-?[xyz]}.*/gi).test(value) ? "Invalid URL or URL doesn't contain {z}, {x} or {y}." : "",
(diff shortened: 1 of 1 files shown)


src/js/imagery/imageryOptions.js
```diff
@@ -63,7 +63,11 @@ export const imageryOptions = [
     {
         type: "xyz",
         label: "XYZ Imagery",
-        params: [{key: "url", display: "XYZ URL"}],
+        params: [{
+            key: "url",
+            display: "XYZ URL",
+            validator: value => !(/https?:\/\/.*{-?[xyz]}\/{-?[xyz]}\/{-?[xyz]}.*/gi).test(value) ? "Invalid URL or URL doesn't contain {z}, {x} or {y}." : "",
```
sirmspencer, line 69: Dont require a slash between xyz.  x={x}&y={y} is also valid for example.

How does this look? `https?:\/\/.*{-?[xyz]}.*{-?[xyz]}.*{-?[xyz]?}.*`

Be more specific with you message.  You are checking for http, x, y, and z.  A generic "invalid url" does not help the user as much.

rfontanarosa: https?:\/\/.*(?=.*{-?x})(?=.*{-?z})(?=.*{-?y}).*
this way ww can check if there is at least one {-?x}, one {-?y} and one {-?z} doesn't matter their position.

PR #1049, opened 2021-02-11
moved rulesViolated and all the validation functions from Collections…

rfontanarosa:
it's not completed: the validators have to be tested but this logic works.


sirmspencer:
src/js/collection.js
816 There is no need to pass ruleError, you can alert from validateAndSetCurrentValue

306 Dont call setCurrentValue if rules are violated.  This does an extra calculation collection.js
306 There is no need to pass ruleError, you can alert from validateAndSetCurrentValue

390 Why did you make a function for this?
392 There is no need to pass ruleError, you can alert from validateAndSetCurrentValue

rfontanarosa:
I choose to pass ruleError to validateAndSetCurrentValue because in collection it is used by a function which do some checks before checking for the ruleError:

`    checkSelection = (sampleIds, ruleError, questionToSet) => {
        if (!this.warnOnNoSamples(this.state.currentPlot)) {
            return false;
        } else if (sampleIds.some(sid => questionToSet.visible.every(vs => vs.id !== sid))) {
            alert("Invalid Selection. Try selecting the question before answering.");
            return false;
        } else if (sampleIds.length === 0) {
            alert("Please select at least one sample before choosing an answer.");
            return false;
        } else if (ruleError) {
            alert(ruleError);
            return false;
        } else {
            return true;
        }
    };`

If these check can be done after ruleError we can remove it from the parameters list.

sirmspencer:
> I choose to pass ruleError to validateAndSetCurrentValue because in collection it is used by a function which do some checks before checking for the ruleError:
> 
> ```
>     checkSelection = (sampleIds, ruleError, questionToSet) => {
>         if (!this.warnOnNoSamples(this.state.currentPlot)) {
>             return false;
>         } else if (sampleIds.some(sid => questionToSet.visible.every(vs => vs.id !== sid))) {
>             alert("Invalid Selection. Try selecting the question before answering.");
>             return false;
>         } else if (sampleIds.length === 0) {
>             alert("Please select at least one sample before choosing an answer.");
>             return false;
>         } else if (ruleError) {
>             alert(ruleError);
>             return false;
>         } else {
>             return true;
>         }
>     };
> ```
> 
> If these checks can be done after ruleError we can remove it from the parameters list.

You can remove the check for ruleError here as well if you are not passing it back up the chain.  I prefer not to pass data back up the chain if its not necessary.  Think about how you had to add the check for ruleError in SurveyQuestions.   You created unnecessary duplicate code.  Also think about how the checks are being done in SurveyCollection.  When we reuse this component it makes more sense to have the error message come from where the error occurs.

PR #1050, opened 2021-02-11
disable deleting last user

gtondapu:
https://github.com/openforis/collect-earth-online/issues/642


sirmspencer:
src/js/review-institution.js
1032-1034 I think there was a little miss comunication between "user" and "member". You can delete the last "member" but not the last user.  Since only admins can delete users they can delete the last "member" and there will still be an "admin" left. I think all of this logic can reduce to `institutionRole === 'not-member' && this.state.institutionUserList.filter(user => user.institutionRole === "admin").length ==1)`
1035 "You cannot delete the last admin of an institution."
1035/1036 Both branches of this logic have side affects.  Use an in if statement. ternaries have an implicit return.  They are to be used when a return value is expected.

PR #1055, opened 2021-02-15
fixed bugs introduced with "move validation PR"

rfontanarosa:


sirmspencer:
428 the anon function needs to take the same amount of inputs as its called with. `getSelectedSampleIds={(q) => [1]}`

PR #1060, opened 2021-02-18
Fix null punning for plotId

sirmspencer:


lambdatronic:
project.sql:
- Line 1219: Is this filter necessary now that all the CTEs are already filtered by _project_id?

PR #1063, opened 2021-02-18
GeoDash - move out addBufer logic

### LINKED ISSUE #1018: /get-plot-sample-geom is called for every widget
Example project https://collect.earth/collection?projectId=19771

It looks like /get-project-plot is called for every widget.  This should be done once when the page mounts.  Line 1124, addBuffer, should be called once return a vector source.  The way the code is currently written, you might need to save the value in state.

rfontanarosa:
fixes #1018


sirmspencer:
This has been superseded by #1064

PR #1064, opened 2021-02-19
Refactor code so that features are only build once

### LINKED ISSUE #1018: /get-plot-sample-geom is called for every widget

sirmspencer:
closes #1018

### DIFF
--- a/src/js/geo-dash.js
+++ b/src/js/geo-dash.js
@@ -111,7 +111,7 @@ class Geodash extends React.Component {
                     const features = [plotJsonObject.geom].concat(sampleGeom)
                         .filter(e => e)
                         .map(geom => new Feature({geometry: mercator.parseGeoJson(geom, true)}));
-                    return new Vector({features: features});
+                    return Promise.resolve(new Vector({features: features}));
                 });
         } else if (plotShape === "square") {
             const pointFeature = new Feature(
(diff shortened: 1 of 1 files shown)


src/js/geo-dash.js
```diff
@@ -75,23 +76,68 @@ class Geodash extends React.Component {
...
+                )
+                )],
+            }));
+        } else if (plotShape === "circle") {
+            return Promise.resolve(new Vector({
+                features: [new Feature(new Circle(
+                    projTransform(JSON.parse(center).coordinates, "EPSG:4326", "EPSG:3857"), radius * 1)
+                )],
+            }));
+        }
```
rfontanarosa, line 138: what does this function return if plotshape is not polygon, circle or square?

rfontanarosa (changes requested):

billyz313 (approved):
Everything looks good, thank you.

src/js/geo-dash.js
billyz313: Is there another option for the plot shape?

PR #1065, opened 2021-02-19
Use byte array encoding to create base64 string

sirmspencer:


rjsheperd:
These changes look good. Especially nice to get rid of that shell command. 👍 

Is there a way to add some testing around this?

PR #1086, opened 2021-02-24
Clean up dump csv / sql

sirmspencer:
The main two accomplishments are
1. Untangle and optimize select queries
2. Have the clojure functions and sql match to reduce processing.

The output should be the same except for the remove of the analysis column which can easily be inferred by looking at the email column.


rjsheperd:
When I pulled this down and downloaded the plots CSV, I noticed that the `center_lat/center_lon/shape/size` data were missing (see screenshot):
ceo-Tyron-Creek-plot-data-2021-03-02_and_ceo-Tyron-Creek-plot-data-2021-03-02__1__and_Collect_Earth_Online

Is there a migration that needs to be run prior?

PR #1104, opened 2021-03-01
[WIP] Autoformat JS files

rjsheperd:
### Purpose
Autoformat existing JS files using `.prettierrc.json` file.

Current rules are:
- Indentation: 4 spaces (do we want 2?)
- No bracket spacing (closer to CLJ format)
- Ternary operations (e.g. `<cond> ? <true-expr> : <false-expr>`) expanded to multiple lines
- `&&`/`||` ops are expanded to multiple lines


sirmspencer:
The PR will need to be re-created off master once development is merged in.

PR #1130, opened 2021-03-10
getArea works with Geometry, sqm to ha fix

### LINKED ISSUE #1129: Statistics widget Area issue
Area is always N/A

rfontanarosa:
fix #1129


billyz313 (approved):
Nice work, ty!

sirmspencer:
If you are dividing by 10000 but then losing decimals, that doesnt seem preferable.  Can you just keep area in sqm so you dont lose precision?

Why does it show a decimal point when there is nothing after the decimal.

PR #1136, opened 2021-03-10
wms url validation

### LINKED ISSUE #1122: WMS Url validator allows empty and not valid url, it only checks for query string

rfontanarosa:
fix #1122


sirmspencer:
77 remove check for http here since you now use urlValidator

PR #1144, opened 2021-03-12
Fix issue with number inputs not initializing correctly.

### LINKED ISSUE #1125: Range survey rule should be inclusive
setting the range from 0-100 is expected to be inclusive. However it is exclusive (0 returns an error)

sirmspencer:
Fixes #1125 

The checks were inclusive already.  The issue was that the numeric input was initialized to the same empty string as a text input.

### DIFF
--- a/src/js/components/SurveyCollection.js
+++ b/src/js/components/SurveyCollection.js
@@ -1,6 +1,6 @@
 import React, {Fragment} from "react";
 
-import {UnicodeIcon, removeEnumerator, intersection} from "../utils/generalUtils";
+import {UnicodeIcon, removeEnumerator, intersection, isNumber} from "../utils/generalUtils";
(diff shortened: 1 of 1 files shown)


rjsheperd:
Seems to be working.

A side note -- I noticed when I created a new project with a numeric input, the placeholder/default value no longer showed up. Instead, it always pre-populated with `0`. Is that the intended result?

Collect_Earth_Online

sirmspencer:
> Seems to be working.
> 
> A side note -- I noticed when I created a new project with a numeric input, the placeholder/default value no longer showed up. Instead, it always pre-populated with `0`. Is that the intended result?
> 
> Collect_Earth_Online

It is not the intended result It appears that number inputs dont actually have placeholders.  This is for the same reason the bug happened.  When a blank value was passed to a numeric input, it shows 0.

PR #1147, opened 2021-03-12
Dont sanitize bad json fields

sirmspencer:


rjsheperd:
This looks good. Are there some reproduction steps that I can run through?

rjsheperd (approved):

sirmspencer:
> This looks good. Are there some reproduction steps that I can run through?

Karen found the issue.  Her comment was that clearly garbage entries resulted in `{}`.  So something like "adfasdfa" should fail now instead of resulting in '{}'.  If you were doing a generative test you can define every possible way that a string can not be json like and it should fail. 

And example of json like string that should be parsed include `"test:y,another:2"` and `"\"key1\":1, \"key2\":\"string\"`.  Either should pass with {} or not.

PR #1148, opened 2021-03-12
Update logic for missing state

sirmspencer:


rjsheperd:
Do you mind sharing what this is related to?

sirmspencer:
AHA. ok, in talking more with the user, it appears that the system is working correctly, however it is actually a problem with the error message. What is happening is that the admin is trying to navigate to a plot previously collected by a user. However the specific plot has not been collected. The error given is "This plot has already been analyzed" when the issue is actually that the plot has not been analyzed.

Image Pasted at 2021-3-12 13-47.png

sirmspencer:
> Do you mind sharing what this is related to?

This is related to Karen's most recent post.  The message to users was incorrectly always the message is if "unanalyzed" plots was selected.

rjsheperd (approved):

PR #1162, opened 2021-03-25
geodash background imagery bug

rfontanarosa:
#fix 1161

### DIFF
--- a/src/js/geo-dash.js
+++ b/src/js/geo-dash.js
@@ -3,7 +3,7 @@ import "../css/geo-dash.css";
 import React from "react";
 import ReactDOM from "react-dom";
 import {mercator} from "./utils/mercator.js";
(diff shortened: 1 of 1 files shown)


sirmspencer:
src/js/geo-dash.js
559, 997, unfortunately your change to make things more clear makes it less clear.  Your function starts with `get` so it should be name after what is being returned. `getInstitutionBaseMap` return the whole base map and not just the id.
999 Your change seems to indicate that the form of the widget has changed.  We unfortunately have not updated older widgets.  You will need to account for both.
```
getInstitutionBaseMapId = basemap => {
    const baseMapId = isNumber(basemap)
        ? basemap
        : parseInt(basemap.id || -1);
    return baseMapId > 0
        ? this.props.imageryList.find(imagery => imagery.id === baseMapId)
        : this.props.imageryList[0];
};
```
Before we merge, check that when basemap is a number that its numeric and not a number string.

PR #1163, opened 2021-03-26
Fully remove unused mailing list

sirmspencer:


rjsheperd:
#### src/js/account.js
101 - Remove `getUserDetails` from `componentDidMount`

#### src/js/register.js
When I tried registering a new user, I can register without checking the `Accept TOS` checkbox:
Screen Shot 2021-03-29 at 12 03 34 PM
Do we need to add an additional check in the backend to present an error?

sirmspencer:
src/js/register.js
When I tried registering a new user, I can register without checking the Accept TOS checkbox:
-- I rebased to get the changes required for the accept TOS warning.

rjsheperd (approved):

PR #1180, opened 2021-04-09
Add image alt tags

rjsheperd:
### Purpose
Add alt texts to images to fix linting error

### DIFF
--- a/src/js/components/PageComponents.js
+++ b/src/js/components/PageComponents.js
@@ -110,7 +110,7 @@ class HelpSlideDialog extends React.Component {
                             </div>
                             <div style={{height: "100%", width: "33vw"}}>
                                 <img
-                                    alt={alt}
+                                    alt={alt || ""}
                                     src={"locale/" + this.props.page + img}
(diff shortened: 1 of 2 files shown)


sirmspencer:
src/js/components/PageComponents.js
113, does this throw a null error?  How about `{alt || ""}` to be safe.
180/254, These are the same logo, but you added different alt text.

src/js/components/PageComponents.js
Thanks for looking up all those links.
330, Just a question, does using a link vs onClick give a different experience to a keyboard user?
390, 395, 407, 413 Put the image attributes on separate lines for consistency.
406, Your URL is for NASA, and the logo is Google.
413, While you are in here can you pick a better size for this icon?  Its way too big on the page.

rjsheperd:
#### src/js/components/PageComponents.js
_330, Just a question, does using a link vs onClick give a different experience to a keyboard user?_ It does because an `<a/>` tag is focus-able, while an element with just an onClick handler cannot be focused by default (you have to add some properties).

sirmspencer:
src/js/components/PageComponents.js
438, Do the rest of the images change size with the screen? If not use PX

PR #1181, opened 2021-04-09
Modal for long survey rule lists

### LINKED ISSUE #1174: Projects where a single question has very many rules shows offscreen and extends the scroll bar
http://local.collect.earth:8080/collection?projectId=9923

image

rjsheperd:
### Purpose
Show a modal window when there are more than 3 survey rules associated with a question.

### Related Issues
Fix #1174

### Screenshots
Tooltip when there are less than 3 rules:
Screen Shot 2021-04-09 at 3 27 57 PM

Tooltip when there are 3 or more rules:
Screen Shot 2021-04-09 at 3 29 41 PM

Modal with all the rules:
Screen Shot 2021-04-09 at 3 28 38 PM


sirmspencer:
83, I think it might make since to move this to the new component where the information is being used.

src/js/components/SurveyRules.js
This is a duplicate name with src/js/project/SurveyRules.js.  Without mentioning whether or not src/js/project/SurveyRules.js is named correctly, I think we can come up with a more specific name here.  Something to note that these rules are only for one question. like `SurveyQuestionsRules`.

11, You dont need a helper function for simple state updates that you are not passing to children.  You can just set the state in each onClick.
20, Is this button for keyboard users?
31, What is aria doing in this case?
35, destructure state like you do with props on 16.  We dont do this consistently, but we need to move towards doing so.
41, Does h5 match our other form headers?  The answer seems to be that we dont have consistency here. h5 matches the quit dialog so its good for now.
51, We have a close SVG Icon
54, There needs to be a scroll area for questions with even more rules than you are testing with.
84, This breaks the current logic.  There should be no icon if there are no rules.

rjsheperd:
_20, Is this button for keyboard users?_ Yes. 
_31, What is aria doing in this case?_ Help deal with screen readers/browsers that won't load images.
_54, There needs to be a scroll area for questions with even more rules than you are testing with._ Bootstrap handles this by allowing the modal to scroll. Should we change that?
_84, Use a ternary expression when all return paths have no side effects._ I prefer not to use ternary expressions for this type of method.

sirmspencer:
83, I think it might make since to move this to the new component where the information is being used. 
-- Do you disagree?

src/js/components/SurveyRules.js
31, What is aria doing in this case? Help deal with screen readers/browsers that won't load images.
-- Lets add screen readers to the agenda at the CEO meeting.

54, There needs to be a scroll area for questions with even more rules than you are testing with. Bootstrap handles this by allowing the modal to scroll. Should we change that?
-- You are correct, this is fine as is.

36 `stopPropogation()` -> `stopPropagation()`
85 For react, returning null will render nothing.  You can use a fun JS shorthand.
```
return rules.length > 0 && (
    rules.length <= 2
        ? <SurveyRulesTooltip rules={rules}/>
        : <SurveyRulesModal rules={rules}/>
);
```

rjsheperd:
> 83, I think it might make since to move this to the new component where the information is being used. Do you disagree?
I agree that this should be moved into a new component. It's pretty nicely encapsulated, so this should be straightfoward.

src/js/components/SurveyRules.js
> 31, What is aria doing in this case? Help deal with screen readers/browsers that won't load images. Lets add screen readers to the agenda at the CEO meeting.
👍

PR #1192, opened 2021-04-14
Add secret mobile route protected by token auth

sirmspencer:
This is for use for very specific external users.

Add new route /simple-collection, which is intended to be a very slimmed down version of the collection page.

Add new authentication that requires the projects token key.  This is to limit access to our external applications as best we can where their use of CEO allows for this narrow definition of a mobile site.

To access /simple-collection, you need the project token.  Query for it and then access the project like `http://local.collect.earth:8080/simple-collection?projectId=18770&tokenKey=4aff465d-d14f-4845-b4dc-2d17efe331e6`


rjsheperd:
Looks good. Tested on Google Chrome with both a logged-in user and an anonymous user (through incognito mode).

rjsheperd (approved):

PR #1197, opened 2021-04-16
Move all rules to modal

### LINKED ISSUE #1190: Eliminate two UX paths for the question rules modal and require click to modal for any number of rules.

rjsheperd:
### Purpose
Move displaying rules to the rules modal to prevent very long tooltips.

### Related Issues
Fix #1190

### Screenshots

2+ Rules
Screen Shot 2021-04-16 at 11 23 15 AM

1 Rule:
Screen Shot 2021-04-16 at 11 23 38 AM

No Rules:
Screen Shot 2021-04-16 at 11 24 22 AM

### DIFF
--- a/src/js/components/SurveyQuestionRules.js
+++ b/src/js/components/SurveyQuestionRules.js
@@ -1,6 +1,7 @@
 import React from "react";
 
 import SvgIcon from "./SvgIcon";
+import {pluralize} from "../utils/generalUtils";
 
 class SurveyRulesModal extends React.Component {
     constructor(props) {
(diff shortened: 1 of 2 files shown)


sirmspencer:
src/js/components/SurveyQuestionRules.js
74 Do you need a wrapper component now?

rjsheperd:
#### src/js/components/SurveyQuestionRules.js
_74 Do you need a wrapper component now?_
Fair point. I would prefer to keep the wrapper since there's another PR coming that moves the `getRulesById` method over to this component, and I prefer how simple the Modal component is.

PR #1198, opened 2021-04-16
Move getRulesById to SurveyQuestionRules

### LINKED ISSUE #1188: Move getRulesById from SurveyCollection to SurveyQuestionRules
See conversation below:

> src/js/components/SurveyCollection.js 83
> 83, I think it might make since to move this to the new component where the information is being used. 
> 83. I think this rule list can be easier to read now that we have more space. Im ok with that being a next PR though. A simple update though could be to replace | with \n.

rjsheperd:
### Purpose
Moves `getRulesById` from SurveyCollection to the SurveyQuestionRules component.

### Related Issues
Fix #1188

### DIFF
--- a/src/js/components/SurveyCollection.js
+++ b/src/js/components/SurveyCollection.js
@@ -496,9 +496,10 @@ class SurveyQuestionTree extends React.Component {
     toggleShowAnswers = () => this.setState({showAnswers: !this.state.showAnswers});
 
     render() {
-        const {hierarchyLabel, surveyNode, surveyQuestions, surveyRules} = this.props;
-        const {selectedQuestion, selectedSampleId} = this.props;
-        const {setSelectedQuestion, validateAndSetCurrentValue} = this.props;
+        const {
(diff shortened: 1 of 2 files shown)


sirmspencer:
499-501, do you need three separate destructuring statements or will 1 work?
FYI, the work to use destructuring is not in the scope of the issue this PR is for.  Next time please create a separate PR for this.
518 / 524, These ids should not be the same.  I dont see id being used here so can we just remove it?
524 `id` is a reserved word and not very descriptive.  Use `surveyNodeId`

sirmspencer (changes requested):
499-501, do you need three separate destructuring statements or will 1 work?
FYI, the work to use destructuring is not in the scope of the issue this PR is for.  Next time please create a separate PR for this.
518 / 524, These ids should not be the same.  I dont see id being used here so can we just remove it?
524 `id` is a reserved word and not very descriptive.  Use `surveyNodeId`

PR #1205, opened 2021-04-16
Move asPercentage, fix linter errors

rjsheperd:
### Purpose
Fix the rest of the linter errors in collection.js

### Related Issues
#1204


sirmspencer:
1425 I think `{userStats && (userStats.plots || "0")}` better represents what we are doing.  If userStats exists, display (userStats.plots with "0" as default value)

PR #1206, opened 2021-04-16
AOI Map - Hold CMD for Mac OS

### LINKED ISSUE #1096: Project > Create > "Hold CMD to Create BBOX" on Mac
### Purpose
As a user on Mac, I have to use CMD to draw the bounding box. Would be nice if the instructions would reflect that

### Screenshot
Collect_Earth_Online

rjsheperd:
### Purpose
When a User is on a Mac, the AOI Map screen instructs the user to "Hold CMD to select an Area of Interest"

### Related Issues
Fix #1096

### DIFF
--- a/src/js/project/AOIMap.js
+++ b/src/js/project/AOIMap.js
@@ -1,6 +1,7 @@
 import React from "react";
 
 import {mercator} from "../utils/mercator";
+import {detectMacOS} from "../utils/generalUtils";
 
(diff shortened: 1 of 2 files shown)


sirmspencer:
162/163 Can we future proof it with `^mac.*` and `macRegex.test(window.navigator.userAgent.toLowerCase())`;

PR #1214, opened 2021-04-26
Format Survey Rules

### LINKED ISSUE #1182: Clean up rule list format now that we have more space in the model
Each rule type can have its own optimal format

rjsheperd:
### Purpose
Now that we have more space for rules inside of the Rules Modal, optimize the format for each rule.

### Steps
- Add truncate(text, max) fn
- Create components for each rule type
- Re-word the rules to refer to Answers
- Truncate long questions
- Italicize questions and and place in quotes
- Put RegEx constrain in `<pre>` tag
- Put answers in Badges

### Related Issues
Fix #1182

### Screenshots
Incompatible Answers
Screen Shot 2021-04-26 at 11 21 01 AM

Text Match (Regex)
Screen Shot 2021-04-26 at 11 21 14 AM

Exact Sums, Min/Max, and Matching Sums
Screen Shot 2021-04-26 at 11 21 48 AM

(NEW) Tooltips for Long Answers
Screen Shot 2021-05-14 at 1 15 00 PM

### DIFF
--- a/src/js/components/SurveyQuestionRules.js
+++ b/src/js/components/SurveyQuestionRules.js
@@ -1,7 +1,78 @@
 import React from "react";
 
 import SvgIcon from "./SvgIcon";
-import {pluralize} from "../utils/generalUtils";
+import {pluralize, truncate} from "../utils/generalUtils";
+
+/**
+ * Helper Components
+ */
+function Badge({children}) {
+    return (
+        <div className="badge badge-light">
+            {children}
+        </div>
+    );
+}
+
+function Pre({children}) {
+    return (
+        <pre style={{display: "inline"}}>
+            {children}
+        </pre>
+    );
+}
+
+function EqualToSumRule({questionsText, validSum}) {
+    const truncjoin = qs => qs.map(q => truncate(q, 15)).join(", ");
+    return (
+        <div className="mb-3">
+            <strong>Sum of Answers</strong>
+            <br/>
+            Answers to <i>{truncjoin(questionsText)}</i> should sum up to <Pre>{validSum}.</Pre>
+        </div>
+    );
+}
+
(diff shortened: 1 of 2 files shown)


rfontanarosa (approved):

sirmspencer:
9/12 The only value allowed as `children` is text.  It may be more clear to pass a prop `badgeText`.
10 If you do the truncating here it will save some repeat code.
11 Add `title` with the full text.
17-23, I think if anything needs pre, its only regex.  Changing font is not easily describable, so I would not use pre for distinction.  In the case of regex if you need pre for it to render correctly, can you update the font so its the same as the rest of the text?  You are putting answers in badges.  It would be more consistent to put all non question values in the badge component.
17-23, I would suggest that using `Pre` is still shadowing pre even with the case difference.  Can you rename, or remove in the case that <pre>  is only needed for regex.
31, 41, 42, 53, 63, 72, 
- Questions vary quite a bit.  Some are full sentences with their own punctuation. Wrap the questions in quotes.
- Truncating at 15 seems too short to me. This is a scrollable area so we have plenty of room to show more of the questions.
- If you truncate text, add the full text as `title`. It may be worth abstracting a question component like you did an answer (as Badge)
30, 40, 52, 62, 71, <br/> is less flexible than other alternatives. It looks like you are just using <br/> to create a new line.  You can use the `d-flex flex-column` bootstrap classes.

sirmspencer:
8, `**/`
24, Just check if the text is an array once. `const textIsArray = Array.isArray(text)`
26, You can use `Array.isArray()` instead of importing lodash
82, does this pre change the font?

rjsheperd:
_82, does this pre change the font?_ Yes, <pre> will use a mono-spaced font that is imported with Bootstrap

PR #1218, opened 2021-04-29
eslint errors and warings have been fixed

rfontanarosa:


sirmspencer:
src/js/widgetLayoutEditor.js
111  This does not actually fix the issue, it just makes the warning go away.  Revert it back so we know to redo that function when we can or use lodash deep clone.
381
```
const maxY = Math.max(...this.state.widgets.map(o => (o.layout.y || 0)))
const yval = maxY > -1 ? maxY + 1 : 0
```
1886 This does not actually fix the issue, it just makes the warning go away.
```
generateLayout = () => _.map(this.state.widgets, (item, i) => {
	const {layout, layout: {w}} = item
	return {...layout, i: i.toString(), minW: 3, w: Math.min(w, 3)}
}
```

PR #1221, opened 2021-04-30
WMS validator: added SRS/CRS check

rfontanarosa:


sirmspencer:
28, this does not account for lowercase crs and srs.  
76-82, this logic is less clear because you are not using isValidJSON.
You can fix both of these by doing a string match on value instead of parsing the JSON.  
```
olProjectionValidator = value => {
	/crs|srs|epsg|wgs/test(value.toLowerCase())
}

PR #1228, opened 2021-05-21
new rule section

rfontanarosa:
## Related Issues
Closes CEO-98

## Submission Checklist
- [ ] Commits include the JIRA issue and the `#review` hashtag (e.g. `CEO-### #review <comment>`)

## Testing

As an Admin, When I create a Project, And I add survey rules, Then I can see it and delete it from the Rules section.

## Screenshots

Untitled

### DIFF
--- a/src/js/components/RulesCollectionModal.js
+++ b/src/js/components/RulesCollectionModal.js
@@ -20,7 +20,7 @@ export default class RulesCollectionModal extends React.Component {
             || (rule.questions && rule.questions.includes(id))
             || (rule.questionSetIds1 && (rule.questionSetIds1.includes(id) || rule.questionSetIds2.includes(id)))
             || (rule.question1 && (rule.question1 === id || rule.question2 === id)))
-        .map(r => (r.questionId && <SurveyRule key={r.id} ruleOptions={r}/>));
+        .map(r => <SurveyRule key={r.id} ruleOptions={r}/>);
 
     render() {
         const {showModal} = this.state;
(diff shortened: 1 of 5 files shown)


sirmspencer:
src/js/project/SurveyRules.js
30, Why did you convert this from a <button> to a <span>? Ask RJ, but I think button is prefered by keyboard users. 
52, use r.type instead of infering from the keys.  See SurveyCardList -> getRulesById.  It looks like I missed that in the review of RJs code in SurveyQuestionRules, where you got this from.  Can you please update both?  Better that that would be to keep all if the individual rule components private and create one RuleDescription component.
```
export function RuleDescription({rule}) {
	return (
		rule.ruleType === "text-match"
		? <TextMatchRule key={r.id} {...r}/>
		: rule.ruleType === "numeric-range"
			? <NumericRangeRule key={r.id} {...r}/>
		...etc
	)
}
```
<div key={r.id} style={{display: "flex", alignItems: "center"}}>
    {inDesignMode && this.removeButton(r.id)}
    <RuleDescription rule={r}/>
</div>
```

rjsheperd:
@rfontanarosa Please use `<button>` or `<a>` tag instead of `<span>` when creating buttons to ensure keyboard navigation is preserved.

rjsheperd:
In testing this, the rules don't seem to show up when I click "Add Rule" (screenshot below):
Screen Shot 2021-07-28 at 5 21 46 PM

PR #1233, opened 2021-06-06
Clean up widgets

sirmspencer:
## Purpose

** Conform widget layout json

Widgets are always an array so there is no longer a need to check.
- See geodash.clj line 20

No widgets with a layout have a:
```
-- missing y
select widget->'layout' from project_widgets
where widget->'layout' is not null 
	and widget->'layout'->'y' is null

-- i that is not already a string
select jsonb_typeof(widget->'layout'->'i') from project_widgets
  where widget->'layout' is not null 
  	and jsonb_typeof(widget->'layout'->'i') <> 'string'
```

Widgets with layouts are cleaned to ensure no left over width or position fields

Widgets with grid row are converted to widgets with layouts

Widgets with width are put one on each row

** Bug fixes
updateAllServerWidgets was called before dashId was set in state so none of the old widgets -> new widgets were save if no other changes were made. updateAllServerWidgets is no longer needed because all of the DB widgets are the same

generateLayout was working on accident as it was called before any widgets were in state, then overwritten.  It was also causing the app to send /update-widget more than once

Use a more specific change comparison because layout adds more fields we don't need in the widget layout settings.  This means all widgets get updated again the first time one changes.

gridcolumn -> layout missed some permutations of the gridcolumn tag.

some widgets had duplicate id / i values, or out of sync id -> i.

## Related Issues
Closes CEO-40

### DIFF
--- a/src/css/geo-dash.css
+++ b/src/css/geo-dash.css
@@ -394,43 +394,18 @@ input:checked+.switchslider:before {
 /*
  * Placeholder dashboard ideas
  */
-.placeholders.row, .placeholdersOld.row {
-    margin-right: auto;
-    margin-left: auto;
-}
-.placeholdersOld{
-    margin-bottom: 30px;
-    text-align: center;
-}
-.placeholdersOld .placeholder{
-    height: 362px;
-}
-.placeholdersOld .minmapwidget {
-    height: calc(100% - 125px);
-}
-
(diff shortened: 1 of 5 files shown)


billyz313 (approved):
Looks good.  Updating all legacy widgets in the db eliminates a lot of excess.

PR #1235, opened 2021-06-07
Update clean up scripts

sirmspencer:
## Purpose
Finalize clean up scripts:
- Use last collected plot instead of project created date
- Ensure that old projects have a created, archived, and collection_start
- The first level of delete for projects leaves meta data (ie project row).  Delete meta data longer

## Related Issues
Closes CEO-147

## Submission Checklist

### DIFF
--- a/src/sql/changes/clean-old-data.sql
+++ b/src/sql/changes/clean-old-data.sql
@@ -67,7 +67,8 @@ FROM (
         availability,
         archived_date,
         EXTRACT(days FROM now() - archived_date) as a_age
-    FROM projects
+    FROM projects, plots
+    WHERE project_uid = project_rid
 ) a
 WHERE availability = 'archived'
     AND a_age > 90;
@@ -94,6 +95,7 @@ FROM (
     LEFT JOIN projects
     ON institution_uid = institution_rid
         AND availability <> 'archived'
+    WHERE archived = FALSE
     GROUP BY institution_uid
     ORDER BY p_cnt desc
 ) a
--- a/src/sql/changes/clean-old-data.sql
+++ b/src/sql/changes/clean-old-data.sql
@@ -15,7 +15,7 @@ DO $$
         WHERE table_schema='ext_tables'
             AND table_type='BASE TABLE'
     ) a
-    WHERE p is null and s is null;
+    WHERE p IS NULL AND s IS NULL;
 
     IF _sql IS NOT NULL THEN
         EXECUTE _sql;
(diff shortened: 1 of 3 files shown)


lambdatronic:
clean-old-data.sql:
- 18: is null -> IS NULL
- 26: plpsql -> PLPGSQL
- 29: `SELECT (SELECT archive_project(project_uid))` ->  `SELECT archive_project(project_uid)`
- 41-44: Align a multi-line CASE clause like so:
```sql
(CASE WHEN MAX(collection_start) IS NULL
      THEN created_date
      ELSE MAX(collection_start)
 END)
```
- 45-48: Fix CASE clause alignment:
```sql
CASE WHEN COUNT(DISTINCT(user_plot_uid)) = 0
     THEN 0
     ELSE COUNT(DISTINCT(user_plot_uid)) / COUNT(DISTINCT(plot_uid))::float
END as complete
```
- 55: A -> a
- 63-74: Don't join the projects and plots tables. Only project
  information is used in the `deep_archive_project` function.
  Therefore, it looks like this code can be simplified as follows:
```sql
-- Deep archive projects
SELECT deep_archive_project(project_uid)
FROM (
    SELECT project_uid, EXTRACT(days FROM now() - archived_date) as a_age
    FROM projects
    WHERE availability = 'archived'
) a
WHERE a_age > 90;
```
76-86: Here's another code block that can be simplified:
```sql
-- Delete deep archive projects
SELECT delete_project(project_uid)
FROM (
    SELECT project_uid, EXTRACT(days FROM now() - archived_date) as a_age
    FROM projects
    WHERE availability = 'archived'
) a
WHERE a_age > 365;
```
92: count -> COUNT
93: from -> FROM
100: `ORDER BY` is unneeded in an automated function like this.

sirmspencer:
63-74: Don't join the projects and plots tables. Only project
information is used in the deep_archive_project function.
Therefore, it looks like this code can be simplified as follows:
-- This is here so that if you run the script again all of the blocks return 0 or no rows.  deep_archive removes plots so the left join is a filter.

92: count -> COUNT
-- On direction from you, we use lowercase for all SQL functions, including built in.

lambdatronic:
clean-old-data.sql:
- 29: `SELECT (SELECT archive_project(project_uid))` ->  `SELECT archive_project(project_uid)`
- 43,48: Align `THEN` below `WHEN`
- 47,49: COUNT -> count
- 58: aa -> a
- 63-74: Don't join the projects and plots tables. Only project
  information is used in the `deep_archive_project` function.
  Therefore, it looks like this code can be simplified as follows:
```sql
-- Deep archive projects
SELECT deep_archive_project(project_uid)
FROM (
    SELECT project_uid, EXTRACT(days FROM now() - archived_date) as a_age
    FROM projects
    WHERE availability = 'archived'
) a
WHERE a_age > 90;
```
- 83: Filter projects table with `WHERE availability = 'archived'`

PR #1238, opened 2021-06-08
Change duplicate imagery title

sirmspencer:
## Purpose
We no longer allow duplicate imagery names.  Go back to older institutions and update the imagery title.

## Related Issues
Closes CEO-28


sirmspencer:
This isnt needed.  We dont allow dup names.  No PK can be set because archived names dont count.

rjsheperd (approved):

PR #1240, opened 2021-06-12
Bulk insert samples

sirmspencer:
## Purpose
Use bulk insert statements for samples to improve time.

Example project with 30 plots and 100 samples went from 40 seconds to .5 seconds.

### DIFF
--- a/src/clj/collect_earth_online/database.clj
+++ b/src/clj/collect_earth_online/database.clj
@@ -99,4 +99,5 @@
                                     :or {fields (keys (first rows))}}]
   (doall (pmap (fn [row-group]
                  (insert-rows! table row-group :fields fields :custom-row custom-row))
-               (pg-partition fields rows))))
+               (pg-partition fields rows)))
+  nil)
--- a/src/clj/collect_earth_online/database.clj
+++ b/src/clj/collect_earth_online/database.clj
@@ -84,18 +84,19 @@
         cat
         rows))
 
-(defn insert-rows!
-  [table rows & {:keys [fields custom-row]}]
+(defn insert-rows! [table rows & {:keys [fields custom-row]
+                                  :or {fields (keys (first rows))}}]
   (let [get-fields (apply juxt fields)]
     (doseq [sm-rows (pg-partition fields rows)]
       (jdbc/execute-one! (jdbc/get-datasource pg-db)
                          (for-insert-multi! table
-                                            (or fields (keys (first rows)))
+                                            fields
                                             (map get-fields sm-rows)
                                             custom-row)
                          {}))))
 
-(defn p-insert-rows! [table rows & {:keys [fields custom-row]}]
+(defn p-insert-rows! [table rows & {:keys [fields custom-row]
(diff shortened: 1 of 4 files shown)


rjsheperd (approved):

lambdatronic:
database.clj:
- 98: Return `nil` from `p-insert-rows!` to match the semantics of `insert-rows!`. Are you concerned at all that this function has an undefined insertion order?

projects.clj:
- These functions all mutate the database and should have names ending with `!`:
  - 140: insert-project-imagery
  - 389: load-external-data
  - 435: create-project-samples
  - 452: create-project-plots
  - 557: create-project
  - 657: reset-collected-samples
  - 687: recreate-samples
  - 742: update-project
  - 860: publish-project
  - 868: close-project
  - 874: archive-project
  - 966: dump-project-aggregate-data
  - 1030: dump-project-raw-data

- 450: There's a subtle bug here in your invocation of `p-insert-rows!`. Within that function, `fields` will be pulled out of the passed-in `samples` maps with `(keys (first samples))`. However, map values aren't guaranteed to have any discernible order in Clojure. Generally, this isn't an issue with `p-insert-rows!` because whatever order the `fields` are pulled out in is the order used in the `get-fields` function created in `insert-rows!`. However, since you specified a `:custom-row` argument to `p-insert-rows!`, you are requiring that the map keys be retrieved in the order `[:plot_rid :sample_geom]` which is not guaranteed. To avoid this potentially breaking in unfortunate and unexpected ways, you must always pass an explicit `:fields` parameter to `p-insert-rows!` whenever you use a `:custom-row` definition. Otherwise, neither are needed.

- 650: Remove print debugging statement: `(println e)`

project.sql:
- 1455: Can we drop `create_project_plot_sample(_plot_id, _sample_geom)`?

sirmspencer:
98: Return nil from p-insert-rows! to match the semantics of insert-rows!. Are you concerned at all that this function has an undefined insertion order?
Interesting point.  Generally in SQL you dont assume order so it should not matter. However, I dont think we are sorting order so it might accidently matter.

Here is where I dont think it matters
1. Samples.  There is no meaning full order to samples or sample IDs.
2. Generated plots.  The user is limited to a number much lower than the 32767 partition size so they will be inserted in one statement
3. Shape files.  These are also a database, I dont think there is a concept of ordering features in a shape files.

So that leaves uploaded CSV files with over 32767 entries.  There could be a situation where the admin intentionally scrambles the plot ids for a random collection order.  I dont necesarily think our current plot navigation techniques lead to predictable behavior here, but I can take the point not to use p-insert-rows! for plots to maintain current behavior until we improve the plot navigation.

650: Remove print debugging statement: (println e)
-- This in intentional.  I want to have unknown errors in the log in production, in case something unexpected goes wrong.  I suppose (log ) is more appropriate here.

1455: Can we drop create_project_plot_sample(_plot_id, _sample_geom)?
-- this is still being used in plots.clj, line 113.  This is for user drawn samples.  It may be possible to use a bulk insert here, but not really necessary with the small handful of samples generated by the user.

PR #1241, opened 2021-06-12
Bulk insert plots

sirmspencer:
## Purpose
Use bulk insert for plots.  Projects that took about 45 seconds to create now take less than 1.

Closes CEO-163

### DIFF
--- a/src/clj/collect_earth_online/db/projects.clj
+++ b/src/clj/collect_earth_online/db/projects.clj
@@ -547,14 +547,17 @@
                              "none"    1.0)
                            200.0
                            50000.0)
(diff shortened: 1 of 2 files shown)


rjsheperd (approved):

lambdatronic:
projects.clj:
- 198: `make-geo-json-point` is unused. However, I suppose we can leave it in this namespace for completeness in case we have a need for it in the future.

- 549: Same subtle map key ordering bug as in PR1240. You need to pass an explicit `:fields` argument to `p-insert-rows!` whenever you specify `:custom-fields`.

project.sql:
- 1122: Can we drop `create_project_plot(_project_id, _center)`?

PR #1251, opened 2021-07-02
CEO-150 Fix flag plots without samples for members

rjsheperd:
## Purpose

Fix bug where an institution member cannot flag a plot that does not have samples.

## Related Issues
Closes CEO-150

## Testing

1. Given I am an institution member, When I flag a plot with a reason, Then the plot is flagged.


sirmspencer:
src/js/collection.js
968, you should not need to check for flagged here anymore since we are letting all flagged plots through.

PR #1252, opened 2021-07-02
CEO-107 Add branded Confirm/Alert modals to Imagery deletion

rjsheperd:
## Purpose

Add branded Confirm/Alert modals to the delete imagery workflow.

## Related Issues
Closes CEO-107

## Testing

1. Given I am an institution admin, And I visit my institution page, And I view the imagery tab, When I click to delete an imagery source, Then I am presented with a confirmation modal.

## Screenshots

Confirm Modal:
Screen Shot 2021-07-02 at 1 13 40 PM

Alert Modal:
Screen Shot 2021-07-02 at 1 14 12 PM


sirmspencer:
src/js/reviewInstitution.js


For example 

So far your changes for a single model look great.  The examples at the top are very helpful.  I want to talk through still using Alert and Confirm modals.  I still we can simplify this code further with just one Modal component.

In looking at the code for both, your consolidated code for Model leaves nothing but a component that calls a component.  Your work for Modal to handle a different set of inputs makes these redundant IMO.

It feels like the code is still a little redundant in reviewInstitution.js (535-550).  If we have one component you wont have to track each type separately. Here is some pseudo code.

Maintain a single state object for your message box.  Since each type or view can have a different set of keys.
```
constructor(props) {
        super(props);
        this.state = {
            ...,
            messageBox: null
        };
    }
```

Display the single Modal (535-550)
```
{this.state.messageBox && (
    <Alert
        {...this.state.messageBox}
        onClose={this.hideMessageBox}
    />
)}
```

As for specific modal types, I think using distinct "show" functions works here.  You have page specific code here in passing 
```
showAlert = ({title, body}) => this.setState({
    type: alert,
    title: title,
    body: body
});

showDeleteConfirm = ({id}) => this.setState({
	type: confirm,
    body: "Are you sure you want to delete this imagery? This is irreversible.",
    title: "Warning: Removing Imagery",
    danger: true,
    closeText: "Cancel",
    onConfirm: () => this.deleteImagery(this.state.deleteImageryId)
});

hideMessageBox = () => this.setState({messageBox: null})

```

rjsheperd:
Thanks for this feedback. While I do see how this this can be written in less code, I also don't know if that helps us that much in this instance. I tried to write `Modal` in such a way that it could be could be built upon into different variations (e.g. `ConfirmModal`, `Alert`), rather than having one component which does it all. For instance, I was hoping to use the `Modal` component to replace the custom modal code inside of `SurveyQuestionRules`. This is dependent on having being able to pass `children` elements into the `Modal` component, rather than simply a `body` text string.

I do appreciate how you have simplified passing the props into the Alert modal using `{...this.state.messageBox}`, and have updated the Alert/ConfirmModals to use that style of passing props.

sirmspencer:
You can do that with the same one MessageBox component:

Display the single Modal (535-550)
```
{this.state.messageBox && (
    <MessageBox
        {...this.state.messageBox}
        onClose={this.hideMessageBox}
    >
       <p>{this.state.messageBox.body}</p>
   </MessageBox>
)}
```

We do the same in Pyregence.  You expanded the code to take a string or a component that renders.

sirmspencer:
src/js/reviewInstitution.js
456, alert -> this.showAlert

PR #1253, opened 2021-07-06
Eliminate the use of separate tables for external data.

sirmspencer:
## Purpose
Eliminate the use of separate tables for external data.

- For plots, reuse the logic we added for samples to allow user drawn.  That is that we dont need to store a center and a geom since center is a geom.  We can then just check for geom type.

- In this code, I use "generators" for each type of plot or sample, and and then uses bulk insert to copy the data to PG.  This should be more flexible moving forward. 
    - Invalid external files fail before being imported into the DB
    - The generators are quicker and potentially reusable for previewing the project in design.
    - Eliminates redundant "recreate-samples"
 
- The DB PK and the visible plot id are now distinguished with obviously different var names.  All plots have a "visible_id".  If none is provided (ie a shapefile 'plotid'), then the number is generated sequentially.
    - A plot ID of 100K+ is not very valuable for users to see
    - All plot types navigate the same
    - Potential to have more flexible ordering, like "scramble" for the grid layout.

- This makes the DB 10% larger, but reduces the backup and restore time by 5x

## Related Issues
Closes CEO-37 CEO-38

## Submission Checklist

## Testing
All creating, collecting, and downloading of project data needs to be tested.

### DIFF
--- a/src/clj/collect_earth_online/db/projects.clj
+++ b/src/clj/collect_earth_online/db/projects.clj
@@ -23,6 +23,15 @@
 (def path-env (System/getenv "PATH"))
 
 ;;;
+;;; Helper function
+;;;
+
+(defn- remove-vector-items [coll & items]
+  (->> coll
+       (remove (set items))
+       (vec)))
+
+;;;
 ;;; Auth functions
 ;;;
 
@@ -457,15 +466,15 @@
                                (update r :visible_id tc/val->int)
                                (dissoc r :lon :lat)))
                            (rest rows))]
-      (if (and (contains? header-keys :lon)
-               (contains? header-keys :lon))
-        [(dissoc header-keys :lon :lat) body]
+      (if (and (some #(= % :lon) header-keys)
+               (some #(= % :lat) header-keys))
+        [(remove-vector-items header-keys [:lon :lat]) body]
         (init-throw (str "The provided "
                          design-type
                          " CSV file must contain a LAT and LON column."))))))
 
 (defmethod get-file-data :default [distribution _ _ _]
-  (throw (str "No such distribution (" distribution ") defined for ollect-earth-online.db.projects/get-file-data")))
+  (throw (str "No such distribution (" distribution ") defined for collect-earth-online.db.projects/get-file-data")))
 
 (defn- check-headers [headers primary-key design-type]
   (let [header-diff     (set/difference (set primary-key) (set headers))
@@ -1008,11 +1017,6 @@
   (when pg-time
     (.format (SimpleDateFormat. "YYYY-MM-dd HH:mm")
              pg-time)))
-
-(defn- remove-vector-items [coll & items]
-  (->> coll
-       (remove (set items))
-       (vec)))
 
 ;;;
 ;;; Dump aggregate
--- a/src/clj/collect_earth_online/db/projects.clj
+++ b/src/clj/collect_earth_online/db/projects.clj
@@ -161,7 +161,7 @@
 (defn- try-catch-throw [try-fn message]
   (try (try-fn)
        (catch Exception e
-         (log (str (ex-message e)))
+         (log (ex-message e))
          (let [causes (conj (:causes (ex-data e) []) message)]
            (throw (ex-info message {:causes causes}))))))
 
@@ -209,7 +209,7 @@
                      "."))))
 
 ;;;
-;;; Spacial plots generator
+;;; Spatial plots generator
 ;;;
 
 ;; Geo JSON
@@ -343,8 +343,8 @@
                              num-plots
                              plot-spacing
                              plot-size]
-  (let [[[left bottom] [top right]] (pu/EPSG:4326->3857 [lon-min lat-min] [lon-max lat-max])
-        [left bottom right top]     (pad-bounds left bottom top right (/ 2.0 plot-size))]
+  (let [[[left bottom] [right top]] (pu/EPSG:4326->3857 [lon-min lat-min] [lon-max lat-max])
+        [left bottom right top]     (pad-bounds left bottom right top (/ 2.0 plot-size))]
     (check-plot-limits (if (= "gridded" plot-distribution)
                          (count-gridded-points left bottom right top plot-spacing)
                          num-plots)
@@ -368,11 +368,6 @@
        (remove #(= folder-name %))
        (filter #(= ext (peek (str/split % #"\."))))
        (first)))
-
-(defn- format-simple
-  "Use any char after % for format."
-  [f-str & args]
-  (apply format (str/replace f-str #"(%[^ ])" "%s") args))
 
 (defn- parse-as-sh-cmd
   "Split string into an array for use with clojure.java.shell/sh."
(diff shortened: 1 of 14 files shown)


src/clj/collect_earth_online/db/projects.clj
```diff
@@ -309,151 +387,224 @@
...
-                          hr))]
-      [headers (type-columns headers) (str/replace body #"," "\t")])))
+    (let [geom-key    (keyword (str design-type "_geom"))
+          header-keys (as-> header-row hr
+                        (str/split hr #",")
+                        (mapv #(-> %
+                                   (str/lower-case)
+                                   (str/replace #"-| " "_")
+                                   (str/replace #"^(x|longitude|long|center_x)$" "lon")
+                                   (str/replace #"^(y|latitude|center_y)$" "lon")
```
rjsheperd, line 450: I think you meant `lat` here

rjsheperd:
Looks good, Matt. 

I tested:
[x] Creating projects with uploading SHP/CSV files for plots and samples
[x] Downloading samples and plots

lambdatronic:
projects.clj:
- 164: `(log (str (ex-message e)))` -> `(log (ex-message e))`

- 212: Spacial -> Spatial (e.g., Spatial Informatics Group)

- 346: Incorrect destructuring order. `pu/EPSG:4326->3857` returns `[[left bottom] [right top]]`

- 347: Incorrect arg order. Call `pad-bounds` as follows: `(pad-bounds left bottom right top (/ 2.0 plot-size))`

- 372: format-simple is not used in this file and is also defined in database.clj

- 377: parse-as-sh-cmd is also defined in build_db.clj and https.clj

- 390: sh-wrapper is also defined in build_db.clj and https.clj

- 429-435: I'm not following the logic here. `header-keys` contains
  one field more than you should get from `body-text` (i.e.,
  :plot_geom or :sample_geom). When you call `(zipmap header-keys r)`,
  the generated map should contain `nil` for either the `:plot_geom`
  or `:sample_geom` field, which means you'll be calling `(str->pg nil
  "geometry")`. Is this your intention, and if so, why? I would have
  expected you to simply rename the auto-generated `geom` field output
  by `shp2pgsql` to either `plot_geom` or `sample_geom` here.

- 461: You called `(dissoc r :lon :lat)` on the body rows but left :lon and :lat in `header-keys`. You should remove these before returning `[header-keys body]`.

- 466: `ceo.db.projects/get-file-data` -> `collect-earth-online.db.projects/get-file-data`

- 468: `primary-key` is confusingly named here because you always pass
  in a vector of keywords. This would be more aptly named
  `required-headers`, in which case you would then want to rename
  `header-diff` to `missing-headers` for greater readability.

  NOTE: `primary-key` should also be renamed to `required-headers` in
  `load-external-data!`, where `check-headers` is called.

- 481: Why are you upper-casing `invalid-headers` in this error
  message but not the missing headers in the error message above? Make
  them match for consistency.

- 492: Wrong `seq?`. You need to use `seq` without a question mark here.

- 511: `(str k2 ": " v2)` -> `(str (name k2) ": " v2)`

- 550: Move `ext-samples` to the next line.

- 604-605: You can use `when-let` to combine these lines:
  ```clojure
  (when-let [bad-plots (seq (map :visible_id (call-sql "plots_missing_samples" project-id)))]
      (init-throw (str "The uploaded plot and sample files do not have correctly overlapping data. "
                       (count bad-plots)
                       " plots have no samples. The first 10 PLOTIDs are: ["
                       (str/join ", " (take 10 bad-plots))
                       "]")))
  ```

- 645: Why not use `(p-insert-rows! "plots" plots)` here?

- 795: Uncommented dead code due to `init-throw` on previous line. You
  should use `#_` to comment out this line until your FIXME is
  implemented. Also, please add further details to this FIXME comment.

sirmspencer:
377: parse-as-sh-cmd is also defined in build_db.clj and https.clj
-- This is also defined in triangulum.  I have not gotten to importing triangulum to ceo.  This will be cleaned up then.

390: sh-wrapper is also defined in build_db.clj and https.clj
-- Similar to above, each file uses a slightly different implementation of that function.  I have an issue to generalize this to triangulum.

429-435: I'm not following the logic here. header-keys contains
one field more than you should get from body-text (i.e.,
:plot_geom or :sample_geom). When you call (zipmap header-keys r),
the generated map should contain nil for either the :plot_geom
or :sample_geom field, which means you'll be calling (str->pg nil "geometry"). Is this your intention, and if so, why? I would have
expected you to simply rename the auto-generated geom field output
by shp2pgsql to either plot_geom or sample_geom here.
-- shp2pgsql uses alter to add the geom column, so its outside the rexeg pattern.  The string parsing does not include geom, so I can just add what ever I want to the end of the vector.
```
CREATE TABLE "bangladesh_hexagons_24oct18" (gid serial,
"global_id" varchar(254),
"country" varchar(254),
"plotid" float8);
ALTER TABLE "bangladesh_hexagons_24oct18" ADD PRIMARY KEY (gid);
SELECT AddGeometryColumn('','bangladesh_hexagons_24oct18','geom','4326','MULTIPOLYGON',2);
```

468: primary-key is confusingly named here because you always pass
in a vector of keywords. This would be more aptly named
required-headers, in which case you would then want to rename
header-diff to missing-headers for greater readability.
--  I think primary key is apt here because order matters.  This is equivalent to a multi column primary key in SQL.  For checking headers order does not matter, but now that we check for duplicates in clj (499) instead of SQL it does.

645: Why not use (p-insert-rows! "plots" plots) here?
-- As you pointed out in a past PR, the order of the plot inserts may matter, and p-insert-rows! does not guarantee order.

lambdatronic:
projects.clj:
- 424-430: Thanks for the ALTER TABLE explanation. I must still be
  missing something though because the :plot_geom or :sample_geom
  field that you conj onto header-keys on line 418 is going to have a
  nil value in the zipmap output on line 427. This ends up calling
  `(str->pg nil "geometry")` on line 428, creating a NULL geometry
  object. Is this right? Where do you end up populating this field,
  and how does it relate to the geom field generated later by
  shp2pgsql in the ALTER TABLE statement?

- 460-465: Two mistakes here. You repeated :lon twice rather than
  including :lon and :lat. Also, you can't call `dissoc` on a vector.
  You need to use your `remove-vector-items` function (line 1012) for
  this. Remember to move `remove-vector-items` above `get-file-data`
  for this to work.
  ```clojure
  (if (and (contains? header-keys :lon)
           (contains? header-keys :lat))
    [(remove-vector-items header-keys [:lon :lat]) body]
    (init-throw (str "The provided "
                     design-type
                     " CSV file must contain a LAT and LON column.")))
  ```

- 468: Almost. `ollect-earth-online.db.projects/get-file-data` -> `collect-earth-online.db.projects/get-file-data`

sirmspencer:
424-430: Thanks for the ALTER TABLE explanation. I must still be
missing something though because the :plot_geom or :sample_geom
field that you conj onto header-keys on line 418 is going to have a
nil value in the zipmap output on line 427. This ends up calling
(str->pg nil "geometry") on line 428, creating a NULL geometry
object. Is this right? Where do you end up populating this field,
and how does it relate to the geom field generated later by
shp2pgsql in the ALTER TABLE statement?
-- The data rows include the geom.  In the header, I exclude geom with regex, then add plot_geom. Then number of headers matches the number of data in the row. 

460-465: Two mistakes here. You repeated :lon twice rather than
including :lon and :lat. Also, you can't call dissoc on a vector.
You need to use your remove-vector-items function (line 1012) for
this. Remember to move remove-vector-items above get-file-data
for this to work.
-- Just FYI, I am removing the keys per your recommendation for completeness. The returned `headers` is only used to check if all of the required headers are there.  Having the extra lat, lon headers dont affect the logic.

PR #1261, opened 2021-07-28
Make proxying optional

sirmspencer:
## Purpose
Make proxying optional for the existing proxiable imagery types.

## Related Issues
Closes CEO-72

## Submission Checklist

## Testing
1. When creating / editing imagery, and select "proxy imagery", then the imagery is saved with that option.
2. When loading a wms source that is not proxied, then the url for the map tiles is not /get-tile
3. When loading a wms source that is proxied, then the url for the map tiles is /get-tile

### DIFF
--- a/src/clj/collect_earth_online/db/imagery.clj
+++ b/src/clj/collect_earth_online/db/imagery.clj
@@ -5,12 +5,13 @@
             [collect-earth-online.views :refer [data-response]]))
 
 (defn- clean-source [source-config]
-  (if (#{"GeoServer" "SecureWatch" "Planet"} (:type source-config))
-    (dissoc source-config [:geoserverParams :accessToken])
-    source-config))
+  (let [image-type (:type source-config)]
+    (cond-> source-config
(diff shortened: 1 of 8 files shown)


rjsheperd:
src/js/reviewInstitution.js
706 - To make the label clickable, you can just wrap the checkbox in the `<label>...</label>` tag, like so:
```
<label>
    <input
        checked={checked}
        className="mr-2"
        onChange={callback}
        type="checkbox"
    />
    {title}
</label>
```

src/js/utils/mercator.js
223 - Did you mean to use `else if` so that is one entire logic tree? Or did you mean to separate that?

sirmspencer:
src/js/utils/mercator.js
223 - Did you mean to use else if so that is one entire logic tree? Or did you mean to separate that?
--- One of the eslint rules we have simplifies this logic tree.  From this form to what you are seeing.
```js
if (isProxied) {
    if (securewatch) {}
} else { 
    if (bing) {
    } else {}
}
```
Im hoping to significantly restructure this logic at a later point.

PR #1270, opened 2021-08-16
Some organization of generators

sirmspencer:
## Purpose
A quick reorganization of the generator code.


rjsheperd:
Great idea to pull those out into a separate "generators" namespace. Approved 👍

PR #1272, opened 2021-08-17
Save external samples for restoring later

sirmspencer:
## Purpose
Save external samples for restoring later.  This is only needed when allow-drawn-samples? is true.

## Related Issues
Closes CEO-185

## Submission Checklist

## Testing
1. Create a project with "allow users to draw samples" and a shp file for samples.
2. As the admin click through some plots and draw samples and delete some of the existing shapes in the plot
3. Publish the project
4. The samples drawn should be gone, and the original shape should exists.


rjsheperd:
Getting this error: 
```
Syntax error reading source at (collect_earth_online/db/projects.clj:732:1).
EOF while reading, starting at line 153
```

rjsheperd:
👍 Tested. 
✅ Approved.

PR #1273, opened 2021-08-18
CEO-95 Enable restricting user-drawn sample geoms

rjsheperd:
## Purpose

Enables restricting the geometries of user-drawn samples. For Centered/Gridded/Random sample distributions, points are always enabled.

## Related Issues
Closes CEO-95

## Submission Checklist

## Testing

1. Given I am an admin, When I create a project, And I enable user-drawn samples, And I restrict sample geometries to Points, Then the collector only allows Points to be drawn.
1. Given I am an admin, When I create a project, And I enable user-drawn samples, And I set sample distribution to Random or Gridded or Centered, Then the point sample geometries are always enabled.

## Screenshots

Project Manager > Sample Designer
Screen Shot 2021-08-18 at 12 52 50 PM

Project Manager > Question Designer > Collector Preview
Screen Shot 2021-08-18 at 12 53 05 PM

Collector
Screen Shot 2021-08-18 at 12 54 20 PM


sirmspencer:
Here are two example of incorrect disable logic
should-not-be-required
no-point

sirmspencer:
45, polygonds -> polygons
42, 97, etc, Sorry for being picky here, what is the difference between project "options" and "settings"?  Something like designSetting, designParameters, etc would be more clear IMO.

src/js/project/SampleDesign.js
62, This function seems unnecessary. when is this null?  It should just be initialized correctly.
66, 77, `const {sampleGeometries} = this.context.projectSettings`
65, This function appears to contain circular logic.  You are setting it to its current value.  Between getSampleGeometry and resetSampleGeometry, I think this is what you are trying to do.
```
resetSampleGeometry = () => {
    this.context.setProjectDetails({projectSettings: {points: true, lines: true, polygons: true}});
};
```
77 `isEnabledSampleGeometry = geometry => this.context.projectSettings?.sampleGeometries[geometry]`
75, This function seems to be tied to the wrong event.  We need to set the point = true when the sample distribution changes, not when the type is toggled.
```
toggleSampleGeometry = geometry => {
    const {sampleGeometries} = this.context.projectSettings;
    this.context.setProjectDetails({
        projectSettings: {...sampleGeometries, [geometry]: !sampleGeometries[geometry]}
    });
};
```
157, Dont use boolean logic for side affects.  I think this is where you should be setting the required point.
```
onChange={e => {
    const newDistributionType = e.target.value
    const pointRequired = ["random", "gridded", "center"].includes(newDistributionType);
    const {points} = projectSettings;
    setProjectDetails({
        allowDrawnSamples: (newDistributionType=== "none") || allowDrawnSamples,
        sampleDistribution: newDistributionType,
        projectSettings: {...projectSettings, points: points || pointRequired}
    });
}}
```
192, I think the indent here looks accidental.  If you only show this section when allowDrawnSamples is enabled, it should be clear that its related.
206, Id suggest hiding or disabling the whole restrict section instead of just the check boxes.
206, 208, options.alwaysEnabled is always truthy for point.  Its a vector. 
208, You should not need to pass the always enabled option.  Set point to enabled on change of distribution type and not here. (see above)
234, with the drawn samples section longer, move these inputs above "Allow users to draw their own samples"

441, when is this null?  You merge in the default options in project.clj before returning

sirmspencer:
45, polygonds -> polygons
42, 97, etc, Sorry for being picky here, what is the difference between project "options" and "settings"?  Something like designSetting, designParameters, etc would be more clear IMO.

src/js/project/SampleDesign.js
7, You should not have to initialize this state here. Initialize the default values for a blank project in projectAdmin.js, line 15.  This allows the project to be completely reset.
62, This function will be unnecessary when projectSettings is initialized.
65, This function appears to contain circular logic.  If you only want to set the initial value, you should check the state instead of just resetting it to itself.  This should also be unnecessary, need to set the point = true when the sample distribution changes, not when the type is toggled (see below).
77 `isEnabledSampleGeometry = geometry => this.context.projectSettings.sampleGeometries[geometry]`
75, This function seems to be tied to the wrong event.  We need to set the point = true when the sample distribution changes, not when the type is toggled.
```
toggleSampleGeometry = geometry => {
    const {sampleGeometries} = this.context.projectSettings;
    this.context.setProjectDetails({
        projectSettings: {...sampleGeometries, [geometry]: !sampleGeometries[geometry]}
    });
};
```
157, Dont use boolean logic for side affects.  I think this is where you should be setting the required point:
```
onChange={e => {
    const newDistributionType = e.target.value
    const pointRequired = ["random", "gridded", "center"].includes(newDistributionType);
    const {points} = projectSettings;
    setProjectDetails({
        allowDrawnSamples: (newDistributionType=== "none") || allowDrawnSamples,
        sampleDistribution: newDistributionType,
        projectSettings: {...projectSettings, points: points || pointRequired}
    });
}}
```
My example does not keep the does not use your more generic `const sampleGeometries`, so feel free to update.
192, I think the indent here looks accidental.  If you only show this section when allowDrawnSamples is enabled, it should be clear that its related.
206, Id suggest hiding or disabling the whole restrict section instead of just the check boxes.
206, 208, options.alwaysEnabled is always truthy for point since it is a vector. 
208, You should not need to pass the always enabled option.  Set point to enabled on change of distribution type and not here.
234, with the drawn samples section longer, move these inputs above "Allow users to draw their own samples"

441, when is this null?  You merge in the default options in project.clj before returning

sirmspencer:
45, polygonds -> polygons

src/js/project/SampleDesign.js
62, you are now mutating an object.  Use the destructuring form I posted.
```
toggleSampleGeometry = geometry => {
    const {sampleGeometries} = this.context.projectSettings;
    this.context.setProjectDetails({
        projectSettings: {...sampleGeometries, [geometry]: !sampleGeometries[geometry]}
    });
};
```
199, this isnt visible when allowDrawnSamples is false. `disabled={options.alwaysEnabled?.includes(sampleDistribution)}`
229-235, with the drawn samples section longer, move these inputs above "Allow users to draw their own samples".  This also keeps the variable length UI at the end of the section.

PR #1276, opened 2021-08-25
Rename assigned to analyzed

sirmspencer:
## Purpose
Rename assigned to analyzed for consistency.


rjsheperd:
Looks good.

L134 - Do you mind removing the comment now that these match?

PR #1285, opened 2021-08-30
CEO-188 Add back end changes for user assignment

sirmspencer:
## Purpose
Add back end changes for user assignment

## Related Issues
Closes CEO-188

## Submission Checklist

## Testing
No testing yet since the UI is not created.  Can manually check if users are assigned by uncommenting 329

### DIFF
--- a/src/clj/collect_earth_online/db/projects.clj
+++ b/src/clj/collect_earth_online/db/projects.clj
@@ -167,14 +167,16 @@
 
 (defn- assign-user-plots
   "Assigns users to plots. The two assignment options are:
-   {:user-method \"equal\"
-    :users       [4581 1 5]}}
+   {:userMethod \"equal\"
+    :users      [4581 1 5]}}
    and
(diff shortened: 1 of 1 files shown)


rjsheperd:
Looking good 👍  

Do we have a way to handle if a user adds or removes a user from the assigned? Or if the percentages of a users change?

sirmspencer:
> Two questions:
> 
> 1. Do we have a way to handle if a user adds or removes a user from the assigned? Or if the percentages of a users change?
   Once users are assigned, they should remain fixed.  However if the user updates the projects, and changes the plot layout, the users would get reassigned.  I think its valid to check for just assignment changes and update them. https://sig-gis.atlassian.net/jira/software/projects/CEO/boards/10/backlog?selectedIssue=CEO-215

> 2. How would you like to handle the `camel->kebab` for the `designSettings` properties?
I will do the integration of the UI for each of these as the UI piece gets merged.

PR #1287, opened 2021-09-09
CEO-186 Add Plot assignment fields

rjsheperd:
## Purpose

Adds fields to the Plot Design page for assigning plots to users.

## Related Issues
Closes CEO-186

## Submission Checklist

## Testing

1. Given I am an admin, When I am creating a project, Then I can assign an equal or percentage of plots to users.

## Screenshots

Screen Shot 2021-09-09 at 11 00 45 AM

### DIFF
--- a/src/clj/collect_earth_online/db/projects.clj
+++ b/src/clj/collect_earth_online/db/projects.clj
@@ -41,7 +41,10 @@
 
 (def default-settings {:sampleGeometries {:points   true
                                           :lines    true
-                                          :polygons true}})
+                                          :polygons true}
+                       :userAssignment   {:userMethod "none"
+                                          :users      []
+                                          :percents   []}})
 
 (defn get-home-projects [{:keys [params]}]
   (data-response (mapv (fn [{:keys [project_id institution_id name description num_plots centroid editable]}]
--- a/src/js/components/Select.js
+++ b/src/js/components/Select.js
@@ -2,16 +2,27 @@ import React from "react";
 
 import {isString} from "../utils/generalUtils";
 
-function Option({option}) {
+function Option({option, valueKey = "value", labelKey = "label"}) {
     const [value, label] = isString(option, "String")
-        ? [option, option] : Array.isArray(option)
-            ? [option[0], option[1]] : [option.value, option.label];
+        ? [option, option]
+        : Array.isArray(option)
+            ? [option[0], option[1]]
+            : [option[valueKey], option[labelKey]];
     return (
         <option key={value} value={value}>{label}</option>
     );
 }
 
-export function Select({disabled, label, id, onChange, value, options}) {
+export function Select({
+    disabled,
+    label,
+    id,
+    onChange,
+    value,
+    options,
+    valueKey,
+    labelKey
+}) {
     return (
         <div className="form-inline">
             <label htmlFor={id}>{label}</label>
@@ -23,10 +34,12 @@ export function Select({disabled, label, id, onChange, value, options}) {
                 value={value}
             >
                 {Array.isArray(options)
-                    /* eslint-disable-next-line react/no-array-index-key */
-                    ? options.map((o, i) => (<Option key={i} option={o}/>))
-                    /* eslint-disable-next-line react/no-array-index-key */
-                    : Object.values(options).map((o, i) => (<Option key={i} option={o}/>))}
+                    ? options.map((o, i) =>
+                        /* eslint-disable-next-line react/no-array-index-key */
+                        <Option key={i} labelKey={labelKey} option={o} valueKey={valueKey}/>)
+                    : Object.values(options).map((o, i) =>
+                        /* eslint-disable-next-line react/no-array-index-key */
+                        <Option key={i} labelKey={labelKey} option={o} valueKey={valueKey}/>)}
             </select>
         </div>
     );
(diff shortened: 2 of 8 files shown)


sirmspencer:
src/js/components/Select.js
5 put each `:?` on its own line.  The linter should then format this consistantly with our other ternaries.
17, 19...etc This object should be in camelCase when we are in JS.  We can parse it as kebab on the back end.
19, Object.assign is a mutating function.  Use the spread operator instead
19, This line is too long.  Use a multiline object definition.
27, Why is user an array
35, 36 splice is not the right function here.  This will only work if you are removing the first user.  it is also mutating.
```
[1, 2, 3, 4, 5, 6, 7].splice(5)
(2) [6, 7]
```
You need slice.  You may want to create us a little removeIndex function.  Slice both ends and rejoin based on the index.  For completeness you could also create a removeItem, that looks for the index based on the item passed in.
32, 81, Your var / param pair is confusing.  Which is user and which is userId?
43, 59, Your var / param pair is confusing.  Which is user and which is userId?
49, Should this be `renderAssignedUsers`?
45, 59, It looks like you already know the index.  Can you pass the index instead of user.id?
99, Move this.props above this.getUserAssignment()
100, Whats going on here?  Positional arrays are not always easy to follow. Why are you not creating an object like on 101?
100/102, You have come up with a few tenses to say the same thing about possible users.  You can reduce the number of vars you need.
```
const assignableUsers = Object.keys(allUsers).map(id => [parseInt(id), allUsers[id]])
	.filter(u => !users.includes(u[0]));
```
You could then use the var name possibleUsers if its more clear as the inverse of assignedUsers.
113, When is userMethod not defined?  This can have hidden bugs from unexpected null.  Initialize userMethod if you need to.
119, You didnt pass in `value` to Select.
129, When is selectedUser not defined?  This can have hidden bugs from unexpected null.  Initialize selectedUser if you need to.

260 This object should be in camelCase when we are in JS.  We can parse it as kebab on the back end.

src/js/project/PlotDesign.js
31, Move this to CreateProjectWizard so that its only called once on load.  I would suggest a promise.all with getTemplateProjects
33, something is wrong here if we need to check between two possible values.  I made a pretty big mistake when designing the context by spreading the project info.  I think that was the only thing on the context initilally and it became too much work at the time to fix it. `institutionId` should be the correct var, it comes in at the page level.
266, move this.state below this.context.  For consistancy the order is destructuring this.context, this.state, this.props, then other definitions

src/js/projectAdmin.js
20 This object should be in camelCase when we are in JS.  We can parse it as kebab on the back end.

src/js/utils/generalUtils.js
183, This is a neat trick to define functions.  However sticking functions on the window obfuscates where the source code is.  You can easily change this to an exportable `function isType(val, typeStr)`

rjsheperd:
src/js/project/AssignPlots.js
_27, Why is user an array_
See below

_100, Whats going on here? Positional arrays are not always easy to follow. Why are you not creating an object like on 101?_
To allow the `<Select />` component to be fairly generic, the options can be one of: an array of values, an array of tuples (e.g. `[value, label]`) or an array of maps (e.g. '{value: <value>, label: <label>}`). I think this allows a fairly wide set of options.

I preferred to just stick with the array of tuples since it seemed easier than having to track two different types of maps.

src/js/project/PlotDesign.js
_33, something is wrong here if we need to check between two possible values. I made a pretty big mistake when designing the context by spreading the project info. I think that was the only thing on the context initially and it became too much work at the time to fix it. institutionId should be the correct var, it comes in at the page level._
This was odd because it was working fine with `institution` on the intial project creation, but later when editing a project the users would not show up. This seemed to fix it.

sirmspencer:
src/js/project/AssignPlots.js
18, Be consistent with the rest of the code.
`const userAssignment = this.getUserAssignment();`
39, 40, I described two functions you could create.  removeItem should take an item as the second param.  Please use removeIdex if you are removing by index.
123, looks like you accidentally deleted some characters.
132, When is selectedUser not defined? This can have hidden bugs from unexpected null. Initialize selectedUser if you need to (which I think you did).

118, I know this is a new one.  Id like to have us migrate to using console.error() for errors.
198, something is wrong here if we need to check between two possible values. I made a pretty big mistake when designing the context by spreading the project info. I think that was the only thing on the context initially and it became too much work at the time to fix it. institutionId should be the correct var, it comes in at the page level.
109, 202, 205, I did like the longer institutionUserList better.

src/js/project/PlotDesign.js
247, I did like the longer institutionUserList better.

rjsheperd:
_132, When is selectedUser not defined? This can have hidden bugs from unexpected null. Initialize selectedUser if you need to (which I think you did)._
This is something that I struggled with, since `allUsers` may not be defined when the initial component is loaded. I had thought it would be best to reset `selectedUser` when `allUsers` does come back from the parent component's AJAX request, but then realized we could just fall back to the first element of `possibleUsers`.

sirmspencer:
src/js/project/AssignPlots.js
11, Can you add comments everywhere you are using positional array information?  Its hard to know what `user[0]` is
11, If you also have users then should this be `selectedUserId`?

sirmspencer:
Dont change variable names when passing props
Incorrect
```
<Component list={userList} />
```

Correct
```
<Component userList={userList} />
```

You changed the format of institutionUserList a few times.  This was unnecessary. You made a pretty generic Select component and stopped just short of being able to use it with institutionUserList as is.

Be careful where you set default values in the code. When you set the value too far down the tree, it can cause unexpected bugs.  Think about initializing the data closer to the trunk.

PR #1299, opened 2021-09-15
Update triangulum and readme to use config for build-db

sirmspencer:
## Purpose
Update triangulum and readme to use config for build-db

https://github.com/sig-gis/triangulum/pull/65


rjsheperd:
README.org
L27 - Remove `:wildfire-camera-api-key` reference

PR #1311, opened 2021-09-23
Filter QA/QC plots on agreement threshold

sirmspencer:
## Purpose
Filter QA/QC plots on agreement threshold for the qaqc navigation mode.

## Related Issues
Closes CEO-221

## Submission Checklist

## Testing
When on the collection page, and my project has qaqc enabled, and more than one plot has been analyzed twice, then clicking next will got to the next plot where the answer to samples to not match with X% under the threshold.
For example if there are 10 plots and 5 are the same and 5 are different, then it should show in the navigation under 50%

### DIFF
--- a/src/clj/collect_earth_online/db/plots.clj
+++ b/src/clj/collect_earth_online/db/plots.clj
@@ -1,13 +1,35 @@
 (ns collect-earth-online.db.plots
   (:import java.sql.Timestamp)
   (:require [clojure.set :as set]
+            [clojure.data.json :refer [read-str]]
             [triangulum.type-conversion :as tc]
             [triangulum.database :refer [call-sql sql-primitive]]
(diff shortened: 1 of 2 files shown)


rjsheperd:
99-100 - Can use `(filterm (fn [[plot-id plots]]` since the `grouped-plots` are an array of `[plot-id plots]` (and skip getting the `plot-id` from the first plot)

sirmspencer:
> src/clj/collect_earth_online/db/plots.clj
> 99-100 - Can use `(filterm (fn [[plot-id plots]]` since the `grouped-plots` are an array of `[plot-id plots]` (and skip getting the `plot-id` from the first plot)

Unfortunately the first plot-id is the visible-id and I need the primary key plot id

PR #1319, opened 2021-09-24
CEO-242 Rename Admin Mode to Review Mode

rjsheperd:
## Purpose

Also changes the default behavior of "Review Mode" to only be enabled by default when a project has been "published."

## Related Issues
Closes CEO-242

## Submission Checklist

## Testing
1. As an admin, When I create a project, And I leave it in Draft Mode, And I go to the project overview page, Then Review Mode is turned off by default.
1. As an admin, When I create a project, And I publish the project, And I go to the project overview page, Then Review Mode is turned on by default.

## Screenshots

Screen Shot 2021-09-24 at 10 41 35 AM

### DIFF
--- a/src/clj/collect_earth_online/db/plots.clj
+++ b/src/clj/collect_earth_online/db/plots.clj
@@ -129,7 +129,7 @@
            :savedAnswers (tc/jsonb->clj saved_answers)})
         (call-sql "select_plot_samples" plot-id user-id)))
 
-(defn- build-collection-plot [plot-info user-id admin-mode?]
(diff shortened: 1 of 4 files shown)


sirmspencer:
src/js/collection.js
205, We need to overwrite any defaults the first time the user goes to the project in published mode. Your original logic is probably closer. `this.setAdminMode(project.isProjectAdmin && (inAdminMode || (inAdminMode !== false && availability === "published)"))`.  Feel free to play around with other logic.
540, we are sort of inverting the process, so I think this should default do analyzed.  Review mode will primarily be used at the end when the admin is reviewing plots.

PR #1327, opened 2021-09-29
CEO-215 Reset plots when assignments are updated

rjsheperd:
## Purpose

Checks the User & QA/QC assignments, and updates the plots when either has been modified.

## Related Issues
Closes CEO-215

## Submission Checklist

## Testing

1. Given I am an admin, When I am updating a project, And I change the User Assignments or Quality Control, Then I am warned that the collected plots will be deleted, And the plots are updated.

### DIFF
--- a/src/clj/collect_earth_online/db/projects.clj
+++ b/src/clj/collect_earth_online/db/projects.clj
@@ -522,13 +522,9 @@
        (same-ring? (exterior-ring (:coordinates geom1))
                    (exterior-ring (:coordinates geom2)))))
 
-(defn modified-assignment? [ds1 ds2]
-  (not= (-> ds1 (tc/jsonb->clj) (:userAssignment))
-        (-> ds2 (tc/jsonb->clj) (:userAssignment))))
-
-(defn modified-qaqc? [ds1 ds2]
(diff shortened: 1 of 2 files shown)


sirmspencer:
525, What design settings to we have that dont need to be checked.?  Use `keep-keys` if only some need to be checked.  In the future we will be adding plot distribution, plot shape, etc to the design setting object so we can just check the whole thing.
```
(defn modified-design? [ds1 ds2]
  (not= (-> ds1 (tc/jsonb->clj) ds1)
        (-> ds2 (tc/jsonb->clj) ds2)))
```

src/js/project/ReviewChanges.js
148, assignmentsUpdated -> plotDesignChanged
149, what settings dont need to be checked?  

In reviewing this code, our simpler designSetting object should probably be `plotDesignSetting` and `sampleDesignSetting` to simplify our update comparisons. This is future work.

rjsheperd:
_What design settings don't need to be checked?_
At this point, we can just check the whole `designSettings` object

_In reviewing this code, our simpler designSetting object should probably be plotDesignSetting and sampleDesignSetting to simplify our update comparisons. This is future work._
Opened [CEO-266](https://sig-gis.atlassian.net/browse/CEO-266)

sirmspencer:
149, Can you just compare the whole object like the back end?

PR #1330, opened 2021-09-30
Use "disagreement" instead of "agreement"

sirmspencer:
## Purpose
Use "disagreement" instead of "agreement" because this is the language Karen uses.

## Related Issues
Closes CEO-252

## Submission Checklist

## Testing
This isnt particularly easy to test without knowing ahead of time what the disagreement should be.  We will need a standard project that is collected for testing.


rjsheperd:
Disagreement screen now shows "undefined" as the answers.
Screen Shot 2021-09-30 at 3 03 20 PM

I think it has to do with the response for `get-plot-disagreement` returning a lazy sequence for `answers`:
Screen Shot 2021-09-30 at 3 04 01 PM

PR #1331, opened 2021-10-01
CEO-213 Clean up Assign/QC layout, Add UserSelect

rjsheperd:
## Purpose

Align all form elements to a grid, rather than use flex alignment. Create a `UserSelect` component to simplify the AssignUsers/QualityControl components.

## Related Issues
Closes CEO-213

## Submission Checklist

## Screenshots

Equal Assignment / Overlap
Screen Shot 2021-10-01 at 10 53 36 AM

Percent Assignment / SMEs
Screen Shot 2021-10-01 at 10 54 29 AM

### DIFF
--- a/src/js/components/Select.js
+++ b/src/js/components/Select.js
@@ -24,10 +24,10 @@ export default function Select({
     labelKey
 }) {
     return (
-        <div className="form-inline">
-            <label htmlFor={id}>{label}</label>
+        <>
+            <label className="col-5" htmlFor={id}>{label}</label>
             <select
-                className="form-control form-control-sm ml-3"
+                className="form-control form-control-sm col-5"
                 disabled={disabled}
                 id={id}
                 onChange={onChange}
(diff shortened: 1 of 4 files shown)


sirmspencer:
src/js/components/Select.js
We have a mix of inline and label over selects.  Are you able to make this more generic for that?
30, I would suggest col-6, which makes the select larger, and still leaves a gap between


src/js/components/UserSelect.js
16, for consistancy with your other logic, use `{possibleUsers.length === 1 ? ["No users to assign."] : possibleUsers}`
30, unfortunately the linter does not adjust the indent automatically for plain text.  Consider a SVG plus.  Your - buttons are not the same size as the plus buttons.  Using SVGs for both might help make them the same.

sirmspencer:
Why are your plus and minus buttons different sizes?

rjsheperd:
_Why are your plus and minus buttons different sizes?_

This is how they look with the current sizes:
Screen Shot 2021-10-04 at 10 22 28 AM

Here's how they look with a consistent size of `1rem`:
Screen Shot 2021-10-04 at 10 23 29 AM

I just thought the remove buttons got too large with a size of `1rem`.

rjsheperd:
Updated screenshots in the PR description ☝️

PR #1334, opened 2021-10-01
CEO-208 Admin save over user plots

rjsheperd:
## Purpose

Enable ability for admins to save over another user's plots.

## Related Issues
Closes CEO-208

## Submission Checklist

## Testing

1. Given I am an admin, When I am reviewing a project, And I am in review mode, And I modify the sample answers, Then the sample answers are overwritten.

### DIFF
--- a/src/clj/collect_earth_online/db/plots.clj
+++ b/src/clj/collect_earth_online/db/plots.clj
@@ -248,20 +248,21 @@
 (defn add-user-samples [{:keys [params]}]
   (let [project-id       (tc/val->int (:projectId params))
         plot-id          (tc/val->int (:plotId params))
-        user-id          (:userId params -1)
+        session-user-id  (:userId params -1)
         plot-user-id     (tc/val->int (:plotUserId params -1))
         review-mode?     (and (tc/val->bool (:inReviewMode params))
                               (pos? plot-user-id)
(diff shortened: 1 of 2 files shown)


sirmspencer:
253, 291, Why is the review mode logic different from the one in get-collection plot?
- Potential question you may be addressing. Should an admin be able to save on to a unanalyzed plot in review mode?
264, 277, 282, Use a binding to avoid repeat calculations.

src/js/collection.js
617, 653, why are you not using currentUserId like in getPlotData?

rjsheperd:
#### src/clj/collect_earth_online/db/plots.clj 
_253, 291, Why is the review mode logic different from the one in get-collection plot?_
I added a check for `(pos? plot-user-id)` since we won't be able to find the plot to save over if that `id` is invalid. Note that we don't need this for the other cases of `review-mode?` since we're modifying existing data.

_Potential question you may be addressing. Should an admin be able to save on to an unanalyzed plot in review mode?_
I don't know if that makes sense for this workflow. Maybe we should circle back with Karen?

sirmspencer:
src/js/collection.js
617, 653, Please dont change variable names when you pass props or call API. Check out getPlotData, where you stayed consistent.

PR #1335, opened 2021-10-01
CEO-253 Warn users flagging a plot will clear answers

rjsheperd:
## Purpose

Adds a confirmation when flagging plots that have existing answers.

## Related Issues
Closes CEO-253

## Submission Checklist

## Testing

1. Given I am a user, When I am collecting, And I have answered some questions, And I flag a plot, Then I am presented with a warning that my answers will be cleared.

## Screenshots

Screen Shot 2021-10-01 at 3 24 00 PM

### DIFF
--- a/src/js/collection.js
+++ b/src/js/collection.js
@@ -596,8 +596,16 @@ class Collection extends React.Component {
         }
     };
 
+    hasAnswers = () =>
+        (Object.values(this.state.userSamples)
+            .reduce((acc, sample) => acc || (Object.keys(sample).length > 0), false)
+        || Object.values(this.state.originalUserSamples)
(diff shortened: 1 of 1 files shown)


sirmspencer:
src/js/collection.js
601, You should be able to use .find instead of reduce.  Its similar to the clojure (some ).
There is a lodash _.isEmpty() if you want.

rjsheperd:
Thanks for that recommendation.

Wanted to note that in testing those changes, I found a bug where if you had answered some q's but not saved, then flagged the plot and hit "cancel" on the confirm alert, it would save the plot with your answers. Separated out the logic now so that doesn't happen.

PR #1359, opened 2021-10-08
Optimization level 1: get all plot samples once for filter

sirmspencer:
## Purpose
Optimize the qaqc nav route by getting all of the samples for the plot at once and grouping in memory.  This can be optimized further, but this is sufficient for now.

## Submission Checklist


rjsheperd:
220 - Remove `time` fn

PR #1378, opened 2021-10-31
Create start creating files per component for geodash editor

sirmspencer:
## Purpose
This is an intermediate step toward re-organizing the code.

## Testing
Everything should work as it has


sirmspencer:
This is not really going to be a relevant intermediate step.

PR #1379, opened 2021-11-02
Update server.clj to allow args to be passed from config.edn

rjsheperd:
## Purpose

Enable the server to take in arguments from `config.edn` to reduce command line headaches.

## Related Issues
Closes TRI-69

## Submission Checklist

## Testing

Server can be started and stopped using the following commands:
- Start: `nohup clj -M:server start &`
- Stop: `clj -M:server stop`

### DIFF
--- a/README.org
+++ b/README.org
@@ -140,11 +140,12 @@ project directory and run:
 npm install
 npm run webpack-dev
 clojure -M:build-db functions
-clojure -M:run-server -m dev [-p 8080] [-P 8443]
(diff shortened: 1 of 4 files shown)


sirmspencer:
src/clj/collect_earth_online/server.clj
44, should the default be true? stop and restart use the repl.  If you think its better for dev to have it off, then add :repo to config.edn so we can set the default to true for servers. (or both)
61, The repl should be loaded after the ssl check on 82, like the rest of the services.
62, Staring -> Starting

PR #1404, opened 2021-12-10
Fix sample count for plots with round shape

sirmspencer:
## Purpose
Fix sample count for plots with round shape

## Testing
1. Create project with round plots and gridded samples
2. Download samples file.
3. Sample IDs should be sequential

### DIFF
--- a/src/clj/collect_earth_online/generators/clj_point.clj
+++ b/src/clj/collect_earth_online/generators/clj_point.clj
@@ -102,7 +102,6 @@
                     visible-offset (-> (dec visible_id)
                                        (* samples-per-plot)
                                        (inc))]
-                (println samples-per-plot)
                 (map-indexed (fn [idx [lon lat]]
                                {:plot_rid    plot_id
                                 :visible_id  (+ idx visible-offset)
--- a/src/clj/collect_earth_online/generators/clj_point.clj
+++ b/src/clj/collect_earth_online/generators/clj_point.clj
@@ -10,9 +10,6 @@
         y-steps (+ (Math/floor (/ y-range spacing)) 1)]
     (* x-steps y-steps)))
 
-(defn- count-gridded-sample-set [plot-size sample-resolution]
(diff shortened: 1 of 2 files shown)


src/clj/collect_earth_online/generators/clj_point.clj
```diff
@@ -105,6 +102,7 @@
                     visible-offset (-> (dec visible_id)
                                        (* samples-per-plot)
                                        (inc))]
+                (println samples-per-plot)
```
Oliver-BE, line 105: Should this be left in?

Oliver-BE:
Approved besides the `println` comment added above. The only thing I noticed is that in the Sample Design page it is possible to input a value for sample spacing such that the text that pops up is green (indicating that it's an acceptable number) but proceeding to the next page still throws a "The sample size limit has been exceeded. Check the Sample Design section for detailed info." alert error. 

Steps to replicate:
1. Spatial Distribution: gridded
1. Plot spacing: 10000
2. Plot Shape: circle
3. Plot diameter: 500m
4. Sample spacing (on the sample design page): 100
5. You now get the error even though the text below sample spacing is green.

sirmspencer:
> Approved besides the `println` comment added above. The only thing I noticed is that in the Sample Design page it is possible to input a value for sample spacing such that the text that pops up is green (indicating that it's an acceptable number) but proceeding to the next page still throws a "The sample size limit has been exceeded. Check the Sample Design section for detailed info." alert error.
> 
> Steps to replicate:
> 
> 1. Spatial Distribution: gridded
> 2. Plot spacing: 10000
> 3. Plot Shape: circle
> 4. Plot diameter: 500m
> 5. Sample spacing (on the sample design page): 100
> 6. You now get the error even though the text below sample spacing is green.

https://sig-gis.atlassian.net/jira/software/projects/CEO/boards/10?selectedIssue=CEO-371

PR #1405, opened 2021-12-10
CEO-348 Pass through visible id so user drawn samples can maintain existing id

sirmspencer:
## Purpose
The uploaded or generated sample visibleId field would turn blank after saving a plot

## Submission Checklist

## Testing
1. Create a project with random or gridded samples. Allow user drawn samples,
2. Download the sample file and take note of the highest original sample id number (e.g. 1000 in a project with 100 plots and 10 samples per plot).
3. Collect, save one plot without drawing custom points and one with drawing custom points.
4. Download the sample data again.
5. Take a look at the sampleid column. All existing sample ids should keep their original id number while any newly drawn points should have ids that start with a number sequentially higher than the highest original sample id from step 2 (e.g. 1001 in our example).

### DIFF
--- a/src/clj/collect_earth_online/db/plots.clj
+++ b/src/clj/collect_earth_online/db/plots.clj
@@ -158,10 +158,11 @@
   (data-response ""))
 
 (defn- prepare-samples-array [plot-id user-id]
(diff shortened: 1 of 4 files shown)


Oliver-BE:
Approved

PR #1411, opened 2021-12-14
CEO-358 FIX User plots are not cleared when samples are rebuilt

sirmspencer:
## Purpose
Fix bug.  We must explicitly remove from user_plots because there is no FK constraint on user_samples

## Submission Checklist

## Testing
1. Create a project
2. Collect in unpublished mode
3. Make changes to only the samples
4. After saving, the plot overview map should be all yellow dots again.

### DIFF
--- a/src/clj/collect_earth_online/db/projects.clj
+++ b/src/clj/collect_earth_online/db/projects.clj
@@ -646,6 +646,7 @@
                 (or (not= samples-per-plot (:samples_per_plot original-project))
                     (not= sample-resolution (:sample_resolution original-project)))))
           (do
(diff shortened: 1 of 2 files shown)


Oliver-BE:
Approved but make this the testing script:

1. Create a project.
2. Collect at least one plot in unpublished mode and make sure to save.
3. Quit out and go the Project Review page. Verify that the plots you collected show up as blue dots in the plot overview map.
4. Now edit the project and on the "Sample Design" tab, make any change to the samples (e.g. change the number of samples per plot).
5. After saving/updating the project, go back to the plot overview map in the Project Review page.
6. All of the plots should now be yellow dots.

PR #1413, opened 2021-12-15
CEO-343 Use distribution to check for fname

sirmspencer:
## Purpose
Template file names were being displayed even if the user changed the distribution type.

## Testing
1. Create a project with shp files
2. Create a new project with that as a template
3. Change distribution type to random
4. On the review page you should not see a file name.

### DIFF
--- a/src/js/project/PlotDesign.js
+++ b/src/js/project/PlotDesign.js
@@ -404,7 +404,7 @@ export function PlotReview() {
                                             </tr>
                                         </>
                                     )}
-                                    {["shp", "csv"].contains(plotDistribution) && (
+                                    {["shp", "csv"].includes(plotDistribution) && (
                                         <tr>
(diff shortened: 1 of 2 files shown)


Oliver-BE:
`.contains` is not a valid function on an array. You want `.includes`.

Oliver-BE:
I locally changed both `.contains` functions to `.includes`, but when trying to create project with the the provided `plot-shape-example.zip` and `sample-shape-example.zip`, I get the following error: "Error creating project: - Plot shp file failed to load."

sirmspencer:
> I locally changed both `.contains` functions to `.includes`, but when trying to create project with the the provided `plot-shape-example.zip` and `sample-shape-example.zip`, I get the following error: "Error creating project: - Plot shp file failed to load."

You find the other bug I was hoping not to fix.  Ill go ahead and fix it.

PR #1417, opened 2021-12-15
WIP Example using svg files without js files with jsx

sirmspencer:
## Purpose
An example showing svg files.  react-svg-loader generates the svg react component for us so we dont have to parse the icon.  It looks like any props passed overwrite the files attributes so we should be able to drop in place.

This should give us a chance to set vertical alignment or not to overwrite bootstrap in flex vs not flex positions.

You may or may not want to keep the SVG icon in a div.  Wed need to look.


Oliver-BE:
This can be closed now that #1425 has been merged into development.

Oliver-BE:
We probably also want to delete this branch.

PR #1422, opened 2021-12-17
CEO-375 Add prop types and default props to SvgIcon.

Oliver-BE:
## Purpose
Adds prop types and default props to the SvgIcon component. Fixes an issue where the default value of `vertical-align: middle` conflicted with padding that was being added to the edit SVG in the review project page.

## Related Issues
Closes CEO-375

## Submission Checklist

## Testing
All SVGs should look as they normally do.

### DIFF
--- a/src/js/components/svg/SvgIcon.js
+++ b/src/js/components/svg/SvgIcon.js
@@ -1,4 +1,5 @@
 import React from "react";
+import PropTypes from "prop-types";
 import {alertIcon} from "./alertIcon";
 import {cancelIcon} from "./cancelIcon";
 import {checkIcon} from "./checkIcon";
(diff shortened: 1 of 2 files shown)


sirmspencer:
src/js/components/svg/SvgIcon.js
91, You dont need to set anything to null here, unless you are specifically checking for null instead of nullish (undefined or null).  nullish is ok here.  IF you dont want to use the nullish condition here, the default would be "unset" or "inherit".

Oliver-BE:
If I don't include a defaultProp for cursor, I get this eslint error: "propType "cursor" is not required, but has no corresponding defaultProps declaration. (react/require-default-props)." I originally had it as null because I was following what the Modal component did:
```javascript
Modal.defaultProps = {
    closeText: "OK",
    confirmText: null,
    onConfirm: null
};
```

PR #1426, opened 2021-12-17
CEO-366 Add Google Analytics.

Oliver-BE:
## Purpose
Add the Google Analytics tag to the site. You can find the scripts that I included on the Google Analytics website by navigating to Admin -> Data Streams -> ceo -> Add new on-age tag -> Global site tag (gtag.js)

## Related Issues
Closes CEO-366

## Submission Checklist

## Testing
This is hard to properly test until deployed to production, but on the Reports Snapshot GA page I can see one user (myself) from using the site on localhost. We might need to figure out how to only track the actual production website since I wouldn't expect my visit to count.

### DIFF
--- a/config.default.edn
+++ b/config.default.edn
@@ -11,4 +11,5 @@
             :base-url "https://my.domain/"}
  :server   {:http-port 8080
             :mode      "prod"
-            :log-dir   "logs"}}
+            :log-dir   "logs"}
+ :ga-id    null}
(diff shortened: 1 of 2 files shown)


sirmspencer:
What do you think about added that tag id to config.edn?

Oliver-BE:
There are now two GA properties set up, one for production and one for development. For each site's `config.edn` file, the proper GA id should be added under the `:ga-id` property.

PR #1432, opened 2021-12-23
Update home header to match geodash

sirmspencer:
## Purpose
A few more updates to make the two headers match.  Also, the original home caret direction was the correct one.

## Screenshots

image
image


Oliver-BE:
home.js
The institution buttons don't have a pointer cursor anymore. Either add back in the class of `btn` on line 524 or add it in manually to the div style.

Small style suggestion:

geoDashHelp.js
Line 39: add `style={{paddingTop: "1rem"}}` to the `h1` element.

PR #1443, opened 2021-12-28
CEO-396 Display uploaded logo filename to the UI 

Oliver-BE:
## Purpose
Adds an `image_name` column to the `institutions` database so that the file name of the logo that's uploaded can be displayed to the user. If no file has been uploaded, then "Choose image..." is displayed to the user.

## Related Issues
Closes CEO-396

## Submission Checklist

## Testing
1. Create an institution.
2. Upload a logo.
3. You should be able to see the file name in the UI.
4. Go to the edit institution page. 
5. The same file name should still be there.
6. Change the logo to something else.
7. The new file name should be there the next time you go to edit the institution.

1. Create an institution without a logo.
2. Go to the edit institution page.
3. The UI should say "Choose image..."

## Screenshots
Screenshot from 2021-12-28 11-20-30
Screenshot from 2021-12-28 11-18-11
Screenshot from 2021-12-28 11-20-10

### DIFF
--- a/src/clj/collect_earth_online/db/institutions.clj
+++ b/src/clj/collect_earth_online/db/institutions.clj
@@ -26,8 +26,9 @@
   (let [institution-id (tc/val->int (:institutionId params))
         user-id        (:userId params -1)]
     (if-let [institution (first (call-sql "select_institution_by_id" institution-id user-id))]
-      (data-response (let [{:keys [name base64_image url description institution_admin]} institution]
+      (data-response (let [{:keys [name image_name base64_image url description institution_admin]} institution]
                        {:name             name
+                        :imageName        image_name
                         :url              url
                         :description      description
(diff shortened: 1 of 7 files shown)


sirmspencer:
src/sql/tables/all.sql
You need a change file in src/sql/changes to match.
22, verify your logic with null on existing projects

src/js/components/InstitutionEditor.js
58, `imageName === ""` and `imageName === null` mean two things here. "" means new institution, or existing institution created with no image after this pr. null means with image or without since all past institution will be set to null.

My suggestion is to not add image_name with a default.  You an create an update query to specifically set values.  set image_name = 'image uploaded' if it has logo_data, and '' if there is no logo data.

Create an issue / PR to eliminate update_institution_logo.  This is legacy from the old way we kept logo images.  You should be able to update the add and update function to pass in the logo data.

PR #1444, opened 2021-12-28
CEO-397 Clean up the user assignment (Dec testing bug fix)

Oliver-BE:
## Purpose
Doesn't allow for a user's role to be updated if the user's role hasn't changed in the input.

## Related Issues
Closes CEO-397

## Submission Checklist

## Testing
1. Go to the Users tab of an institution. 
2. Change the role of a user to a different role.
3. Change the role of that user back to its original value.
4. Try to click the Update button for that user.
5. An alert warning you that you must change the user's role should pop up when clicking on it.

## Screenshots
Screenshot from 2021-12-28 15-21-08

### DIFF
--- a/src/js/reviewInstitution.js
+++ b/src/js/reviewInstitution.js
@@ -1322,8 +1322,7 @@ class User extends React.Component {
     constructor(props) {
         super(props);
         this.state = {
-            userRole: props.user.institutionRole,
-            didRoleUpdate: false
+            userRole: props.user.institutionRole
         };
     }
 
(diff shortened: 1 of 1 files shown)


sirmspencer:
src/js/reviewInstitution.js
1326, 1361, 1366, dont save derived values in state.
1257-1369, revert this.  Also for reference, this is overly complex, you could have done
```
this.setState({
    userRole: e.target.value,
    didRoleUpdate: this.props.user.institutionRole !== e.target.value
});
```
In this form it might be a litter easier to see that you are saving derived values in state.  userRole will be equal to e.target.value and this.props.user.institutionRole is a constant value.
1380 this.setState.userRole !== this.props.user.institutionRole
1381, only disabled group is changing. Use string literals to only have that part vary.
1384, this.setState.userRole !== this.props.user.institutionRole
1384, You dont need to check if the button is disabled

1380-1387, Please use destructuring for props and state. 

In general I prefer to not disable items, but instead alter the user as to why.  So instead of disabling at all, if this.setState.userRole === this.props.user.institutionRole, alert the user

PR #1448, opened 2021-12-29
Use "id" for id keys

sirmspencer:
## Purpose
We reference "parentQuestion" as a question object, which conflicts with the key "parentQuestion" in the question object. Using "parentQuestionId" for the key will clear confusion.

## Testing
All survey pieces should work.

### DIFF
--- a/src/js/collection.js
+++ b/src/js/collection.js
@@ -455,7 +455,7 @@ class Collection extends React.Component {
             : {},
         selectedQuestionId: Number(findObject(
             this.state.currentProject.surveyQuestions,
-            ([_id, sq]) => sq.parentQuestion === -1
+            ([_id, sq]) => sq.parentQuestionId === -1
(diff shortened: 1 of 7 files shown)


Oliver-BE:
This does not work with `dev_data`. Issues: trying to use the default "Test Project" as a template breaks the website when you navigate to the Survey Questions page in the Project Wizard. Going to the Project Review page for the default "Test Project" doesn't show any of the Survey Questions, just the Survey Rules.

sirmspencer:
> This does not work with `dev_data`. Issues: trying to use the default "Test Project" as a template breaks the website when you navigate to the Survey Questions page in the Project Wizard. Going to the Project Review page for the default "Test Project" doesn't show any of the Survey Questions, just the Survey Rules.

We will work on the dev data at the end of the data changes

PR #1467, opened 2022-01-12
Some small bug fixes with Rules.

Oliver-BE:
## Purpose
Fixes a few small bugs and makes two small UI improvements. 
Bug fixes:
1. The length of an object was being calculated in the wrong way.
2. Two error messages were not being calculated correctly.
3. New rules were being added with an ID of null, which caused a number of issues.

UI enhancements:
1. Added a pointer cursor to the different steps in the Project Wizard.
2. Added a separate section for Rules in Project Review (in anticipation of the Survey Rule Card PR).

## Submission Checklist

## Testing
Clicking on the edit button of the Rules section in the Project Review page takes you right to the Rules step in the Project Wizard. 

Adding rules works as expected.

## Screenshots
Screenshot from 2022-01-12 14-31-00

### DIFF
--- a/src/js/project/CreateProjectWizard.js
+++ b/src/js/project/CreateProjectWizard.js
@@ -495,6 +495,7 @@ export default class CreateProjectWizard extends React.Component {
                     style={{
                         borderRadius: "50%",
                         backgroundColor: stepColor,
+                        cursor: "pointer",
                         height: "2.5rem",
                         width: "2.5rem",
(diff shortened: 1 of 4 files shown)


sirmspencer:
src/js/project/ReviewForm.js
103, your two headers are at different levels in the HTML. 

src/js/project/SurveyRules.js
55, you can move this to the line before and remove the extra set of ()

PR #1474, opened 2022-01-14
CEO-410 Replace Rule tooltip with SurveyRule

Oliver-BE:
## Purpose
Replaces the tooltip content from each Survey Question Card and replaces it with a SurveyRule component. Small clean up with the logic for checking if a specific question includes a rule.
  
## Related Issues
Closes CEO-410

## Submission Checklist

## Testing
When looking in the "Rules" section of a survey card and hovering over a Rule, you should see the Rule Card as a tooltip.

## Screenshots

Screenshot from 2022-01-14 14-45-35

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -458,6 +458,9 @@ a:hover {
     width: 500px;
     z-index: 1;
 }
+.tooltip_content.survey_rule::after {
+  border-color: transparent transparent white transparent;
+}
 
 /*************/
 /* Home Page */
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -443,6 +443,21 @@ a:hover {
     padding: 5px 5px 5px 25px;
     width: 300px;
 }
+.tooltip_content.survey_rule {
+    visibility: hidden;
+    background-color: transparent;
+    border-radius: 0;
+    color: #000000;
+    left: 0;
+    font-weight: normal;
+    max-width: max-content;
+    text-align: start;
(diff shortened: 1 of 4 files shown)


sirmspencer:
185, why did you convert to a set here?  That seems like a extra type conversion since we are only looking up once.
194, SurveyRule has a div as its highest element.  Ill bet you dont need to nest it in another div.

The little arrow that is created with ::after is still black for your white tool tip.  I was originally thinking that we would just include the body of the rule since the title and number are already in the ul.  If you can show the remove button in the tool tip that would be a great reason to show the whole card.  Id guess that you could move the jsx for the button to the SurveyRule component and only pass in a remove rule function.

Oliver-BE:
"185, why did you convert to a set here? That seems like a extra type conversion since we are only looking up once."

I converted the array to a set here because this gives us a code speed up. Before, we were iterating over each rule, creating an  array from all of the potential key names for the `questionId`, and calling `.includes` on that array. If there are `n` rules and `m` questionIds, then that takes O(nm^2) time since for each rule we create an array (which takes O(m)) and call a lookup on that array (also O(m)). If we convert to a set, then it only takes O(nm) since we have a constant time lookup of a specific `questionId` and creating the set from an array is also just O(m).

Oliver-BE:
Screenshot from 2022-01-18 13-48-40

sirmspencer:
`removeFn` is ambiguous. -> `removeRule`

PR #1476, opened 2022-01-19
CEO-421 Fixing bugs with Rules when reordering questions.

Oliver-BE:
## Purpose
Fixes an old bug where swapping the order of survey questions in the Project Wizards wouldn't swap the rules properly. This would break any existing rules if a survey question was reordered, since it would always point to the same question ID/index. This PR adds code such that the question IDs inside of the `surveyRules` are updated properly every time a question is reordered in the Project Wizard.

## Related Issues
Closes CEO-421

## Submission Checklist

## Testing
1. Create a new project with many different survey questions.
2. Create at least one rule of each type.
3. In the Project Wizard, use the up and down buttons to reorder the questions a bunch of times.
4. Each rule should stay attached to the question it is meant for.

### DIFF
--- a/src/js/project/SurveyCardList.js
+++ b/src/js/project/SurveyCardList.js
@@ -40,16 +40,64 @@ class SurveyCard extends React.Component {
         : val === swapVal ? checkVal
             : val);
 
+    swapIdArray = (arr, checkVal, swapVal) => {
+        const newArr = [...arr];
+        if (newArr.includes(checkVal) && newArr.includes(swapVal)) {
+            return newArr;
+        } else if (newArr.includes(checkVal)) {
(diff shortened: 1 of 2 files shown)


sirmspencer:
44, instead of creating a new array to mutate, create a function in sequence.js to replace a value at an index. Use removeAtIndex as inspiration.
56, Gary prefers explicit else instead of implicit else at the end of the function. You have two paths that return the unmodified array.  You can update your logic in your if block to simplify.
60/66 you dont need all ids here.  You already will need to have a if statement for each rule type so you can check each type specifically.
68, 72, 77, 81, Use rule.type for your check instead of derived values.
65, you create newRule and then mutate it.  Instead return the new value from each if block.

PR #1480, opened 2022-01-20
Fix plot column order on plot output

sirmspencer:
## Purpose
This fixes a bug reported here. https://mattermost.sig-gis.com/sig/pl/m9656pcdstyuzn33zyadwzp34h


Oliver-BE:
It looks like you changed the column order but not the types of the column. I'm not sure if this was intentional or not.

sirmspencer:
> It looks like you changed the column order but not the types of the columns you switched. I'm not sure if this was intentional or not.

It was.  Only the names were wrong.  The types still match whats returned below.

PR #1481, opened 2022-01-21
CEO-420 Fix issue with creating parent answer

sirmspencer:
## Purpose
Small bug with parent question answers not showing.

## Testing
1. Create a new project.
2. Create a question with two answers (I used a button - text component)
3. Go to create a new question and select "Parent Question" to be the question you just created.
4. Click the Parent Answer drop down.

### DIFF
--- a/src/js/collection.js
+++ b/src/js/collection.js
@@ -791,7 +791,7 @@ class Collection extends React.Component {
             .forEach(feature => {
                 const sampleId = feature.get("sampleId");
                 const userAnswer = _.get(this.state, ["userSamples", sampleId, selectedQuestionId, "answerId"], -1);
-                const color = componentType === "input" && userAnswer > 0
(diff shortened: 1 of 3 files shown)


Oliver-BE:
This fixes the issue of not having any options in the Parent Answer drop down, but I discovered a new bug with this. Steps to replicate:
1. Create a new project.
2. Create a question of type dropdown text with at least two answers.
3. Create a new question (still of type dropdown text) and select "Parent Question" to be the question you just created. Add at least one answer.
4. In the survey question tree on the right hand side of the page, try to select any of the dropdown answers from your first question.
5. The page crashes and the following error is thrown:  
```
Uncaught TypeError: Cannot read properties of undefined (reading 'parentQuestionId')
    at SurveyCollection.componentDidUpdate (SurveyCollection.js?34e0:35:1)
```

PR #1490, opened 2022-01-24
CEO-100 Allow users to select multiple parent answers

sirmspencer:
## Purpose
Allow users to select multiple parent answers

## Testing
1. Create a project
2. Add a question with many answers
3. Add a child question and select some but not all of the answers
4. The survey preview and the collection survey should only show the child question for the selected parent answers

## Screenshots
image
image


sirmspencer:
Merge dev and update dev data

PR #1494, opened 2022-01-25
CEO-368 Show running total for percent plot assignments.

Oliver-BE:
## Purpose
Shows the running total percent of plots assigned in the Plot Design. The text is red if the percents don't add up to 100%, and green if they do. 

## Related Issues
Closes CEO-368

## Submission Checklist

## Testing
1. Create a project.
2. Assign at least two users using the "Percentage of plots" method.
3. Mess around with their percentages.
4. The running total percent should be visible at the bottom of the last user and the text should only be green once the total percent is 100.

## Screenshots
Screenshot from 2022-01-25 10-18-18
Screenshot from 2022-01-25 10-18-35


sirmspencer:
src/js/project/AssignPlots.js
127, Create a sumArray function in sequence.js. (prev, current) -> (acc, cur). acc is accumulator.  You are not get the previous value, you are getting the accumulated value.

PR #1500, opened 2022-01-26
CEO-432 Filter public projects from templates by default

sirmspencer:
## Purpose
Filter public projects from templates by default.  The template list has too many useless public projects

## Testing
1. Import live data so there is enough to use
2. Have / create at least one project in a institution
3. Go to the create project page
4. The template dropdown should only show institution projects
5. Check the Show Public Projects option
6. The template dropdown should show public projects

## Screenshots
image


Oliver-BE:
I'm only able to see projects when selecting the "Show Public Projects" checkbox. When the checkbox is not clicked (and I'm creating a project in an institution that has institution projects), I don't see anything in the drop-down.

PR #1521, opened 2022-02-09
CEO-21 Allow users to shuffle plot order for gridded and random

sirmspencer:
## Purpose
This is primarily for gridded plots where some users what the collection order to be random.  For random strata, the data is currently collected in order of strata, random in each.  This will allow all strata to be randomized together.

## Submission Checklist

## Testing
1. Create a project with gridded plots
2. Check the Shuffle Plot Order box
3. Collect to verify that the plots are in a random order. That is, when you zoom out while collecting on each of the first few plots, they aren't directly next to each other. You can also download the plot data CSV to verify that the lon and lat aren't in order.

## Screenshots
image


src/sql/changes/update_2020-02-10_allow_shuffle_plots.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE projects ADD COLUMN shuffle_plots boolean;
```
Oliver-BE, line 1: This file has the wrong year - it should be `update_2022-02-10_allow_shuffle_plots.sql`

PR #1522, opened 2022-02-10
Fix some issues with gee institution imagery

sirmspencer:
## Purpose
Users reported issues with loading GEE institution imagery.  The main issue was no default reducer since the UI does not take that as an input.

## Submission Checklist

## Testing
1. Create an institution imagery type GEE Imagery Collection
2. Add it to a project
3. Collect with that project
4. Select the imagery type

* Note, I was still getting different errors.  I am not sure if that is on our side or the settings the user gave.  Error should not be of type "none has no function lower"

### DIFF
--- a/src/py/gee/routes.py
+++ b/src/py/gee/routes.py
@@ -49,7 +49,7 @@ def imageCollection(requestDict):
     values = imageCollectionToMapId(
         getDefault(requestDict, 'assetId', None),
         getDefault(requestDict, 'visParams', None),
-        getDefault(requestDict, 'reducer', None),
+        getDefault(requestDict, 'reducer', "Mean"),
         getDefault(requestDict, 'startDate', None),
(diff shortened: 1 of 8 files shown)


src/py/gee/routes.py
```diff
@@ -49,7 +49,7 @@ def imageCollection(requestDict):
     values = imageCollectionToMapId(
         getDefault(requestDict, 'assetId', None),
         getDefault(requestDict, 'visParams', None),
-        getDefault(requestDict, 'reducer', None),
+        getDefault(requestDict, 'reducer', "Mean"),
```
Oliver-BE, line 52: Do we use single quotes or double quotes for strings in Python stylistically? We should probably choose one and stick with it.

sirmspencer: Good question.  The previous author was very inconsistent. Ill at least make it consistent within this function.

PR #1526, opened 2022-02-15
CEO-443 Add a dropdown to select between file upload/manual selection for Boundary Coordinates

sifbiri:
## Purpose

Make use of dropdown/selector to choose between input methods of plot boundary, depending on the choice the map can be active (when needed) or non-active with color cues and moving its instruction text position to bottom. 

## Submission Checklist

## Testing

1. In project Wizard, in Plot Design 
2. Select: Upload Boundary File 
3. The map section should change Color, loses interactivity, instruction message disappears
4. Select: Input Boundary Coordinates 
5. The map section should become normal, active with instruction message

## Screenshots

image
image

### DIFF
--- a/src/js/project/AOIMap.js
+++ b/src/js/project/AOIMap.js
@@ -117,9 +117,10 @@ export default class AOIMap extends React.Component {
 
     render() {
         return (
-            <div id="project-map" style={{height: "25rem", width: "100%"}}>
+            <div style={{height: "25rem", width: "100%"}}>
+                <div id="project-map" style={{height: "25rem", width: "100%"}}> </div>
                 {this.props.canDrag && (
-                    <div className="col small text-center mb-2">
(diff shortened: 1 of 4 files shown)


sirmspencer:
110/517, dont name state and class methods the same.

src/js/project/PlotDesign.js
240, renderAOISelector is not being used
247, Dont create a component inside another component.  If this needs to be a component move it outside. If you want it to stay inside PlotDesign, move state to the PlotDesign level, and use a regular ES6 function. `renderAOISelector`
248, your state variables are a bit vague.  Please use more specific names.
267, Use padding or margin instead of br.

sirmspencer:
56, use the same name for props and the value passed.
110, This is a nice and descriptive var.  I think you can get away with `isHelpInactive`
517, your function does not deactivate help component, it sets a value to isHelpInactive. Use `setIsHelpInactive`

PR #1534, opened 2022-02-23
CEO-455 Use global color vars for the CSS

sifbiri:
## Purpose
Create global color vars instead of using hex values everywhere
## Submission Checklist

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -10,6 +10,9 @@
     --yellow-on-hover:      #ecbc0d;
     --lightgray:            #eceeef;
     --lightgray-on-hover:   #dadcdf;
+    --gray:                  #808080;
+    --darkgray:             #5e5f60;
+    --darkgray-on-hover:    #3a3b3c;
 }
 
 /***********/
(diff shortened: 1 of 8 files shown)


sirmspencer:
Great start!  There are other places in the rest of the code that reference the hex.  Can you update these as well?

sirmspencer:
src/css/custom.css
I had "hide white space changes" set in my last review.  I see now that you changes the indent from 4 to 2. We can talk about doing that at a later point, but it should not be a part of this PR.

469, I was expecting a change here from PR 1532.  It looks like your a few commits behind (I think you based off of 1533).  You can pull the latest development and rebase to see the changes from 1532.

sirmspencer:
475, should the yellow be --yellow?
475, Lets define a var for gray so we can modify it in the future if we need to.  Its also possible the default gray is different for each computer if we dont define it specifically.
475, I think it would be cleaner to set stepColor to the color only and combine with var where used.  `const stepColor = isSelected ? "--lightgreen" : stepComplete ? "--yellow" : "--gray";` or just reduce some redundant code: `const stepColor = 'var(${isSelected ? "--lightgreen" : stepComplete ? "--yellow" : "--gray"})';` (note that I cant type back ticks into github the single quote should be a back tick)

sifbiri:
I don't 

475, Yeah...I corrected it before but somehow slipped with cherry-picking commits and all.
475, Yes, I was debating that in my head. Makes sense. 
475, What if we add a small utility function that decorates the color string to a css var that we can also use elsewhere?

sirmspencer:
This is a little simpler version of my comment above 'var(--${isSelected ? "lightgreen" : stepComplete ? "yellow" : "gray"})';`

PR #1535, opened 2022-02-25
CEO-458 The help title has no color if upload boundary file is selected

sifbiri:
## Purpose

Disable the Map Help Panel when not needed to collect input and change its colors to ligth gray to signal inactivity to users. 
## Submission Checklist

### DIFF
--- a/src/js/project/CreateProjectWizard.js
+++ b/src/js/project/CreateProjectWizard.js
@@ -525,7 +525,7 @@ export default class CreateProjectWizard extends React.Component {
         const steps = this.getSteps();
         const {description, StepComponent, helpDescription, StepHelpComponent} = steps[this.context.wizardStep];
         const isLast = last(Object.keys(steps)) === this.context.wizardStep;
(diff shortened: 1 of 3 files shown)


sirmspencer:
558, This is incorrect.  You are getting false results because isMapInactive is always undefined (the value would be this.state.isMapInactive). Just set the class to bg-lightgreen since there is no reason it should go to bg-light

sifbiri:
isMapInactive is defined as a const in the method render. It is local variable that depends on both the state and context (current step). When the file input boundary is selected, the map is not needed and becomes disabled. I think from user perspective would be nice to signal it by changing color (gray). Just like when a button is inactive it becomes gray. 
Though maybe changing the const name to a different one, so it doesn't create confusion.

sirmspencer:
I checked again, when you upload a file for the boundary, it shows on the map.  Im back to thinking the title doesnt need to change color (ie disabled).

image

sirmspencer:
I am opening a new PR to make some changes.  Your code is fine, but I think we want to do this a little different.

PR #1555, opened 2022-03-10
CEO-480 Dont show "Rules:" when there are none

sifbiri:
-------

## Purpose

Dont show "Rules:" when there are none

## Submission Checklist


sirmspencer:
src/js/survey/SurveyDesignQuestion.js
58, 121, You are null checking twice `surveyRules && ...`.  What do you think of a ternary on 58 so that filteredRules is [] when surveyRules is null?

PR #1563, opened 2022-03-21
CEO-493 import json, fix type checking 

jdilger:
Add missing json import. Check value type with isinstance() rather than a string. Checking a dictionary against a string returns false and without json imported every time safeParseJSON is called an exception is thrown - thus resetting any visParams to an empty dictionary. This should fix CEO-493 unless there are other issues. I haven't had a chance to lest locally yet. 

## Purpose
Need to import json to use json.load()
Checking instance type using isinstance(). Checking type with a sting returns false (python 3.8.12)

## Related Issues
CEO-493

## Submission Checklist
- [ ] Included Jira issue in the PR title (e.g. `CEO-### <title>`)

## Module(s) Impacted

Geodash 

## Testing

1. Given..., When ..., Then ....

### DIFF
--- a/src/py/gee/utils.py
+++ b/src/py/gee/utils.py
@@ -3,6 +3,7 @@ import os
 import ee
 import math
 import sys
+import json
 from ee.ee_exception import EEException
 from gee.inputs import getLandsat, getS1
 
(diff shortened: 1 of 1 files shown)


sirmspencer:
This is superseded by #1563

PR #1566, opened 2022-03-25
Fix reducer renaming bands

jdilger:
## Purpose
col.reduce(ee.Reducer) renames all bands to bandName_reducerName which causes a mismatch between the band names and image visualization json.

Fix: Get the band names before the reduction and their order before reducing the image collection. After the collection has been reduced to an image select the bands in the proper order and rename them to the original band names. 

Notes: No Jira issue that I can find. Not tested on local dev site, but python GEE code tested and works. 

## Related Issues
Closes CEO-###

## Submission Checklist
- [ ] Included Jira issue in the PR title (e.g. `CEO-### <title>`)

## Module(s) Impacted
Imagery

## Testing

1. Given..., When ..., Then ....


sirmspencer:
superseded by #1567

PR #1587, opened 2022-04-22
CEO-487 Statistics widget not reporting elevation and other stats.

jdilger:
## Purpose
Added import utils to __init__ to allow for autocompletion in VS Code. This makes development and testing easier.

The statistics widget was not reporting elevation statistics for small plots (e.g. smaller than 100m2). The scale for calculating the statistics was set too high to compute anything for these areas. The solution is to have a smaller scale and add the bestEffort parameter.  bestEffort will adjust the scale of the computation to ensure it does not fail. 

## Related Issues
Closes CEO-487

## Submission Checklist

## Module(s) Impacted
GeoDash

## Testing
Test small plots return statistics.
1. Create a new project .
2. In Plot Design - Add a random distribution of 2 plots and set plot width to 1m.
3. In Sample Design - choose Center for sample distribution.
4. Create project.
5. Add a Statists GeoDash widget.
6. Navigate to the first plots GeoDash widget in collection mode and inspect.

Desired outcome: The widget should display a numerical value for 'Elevation'. 'Elevation should not be 'N/A - N/A'

Test large plots return statistics.
1. Edit previous project.
2. In Plot Design - Update plot width to 10,000m.
3. Confirm edits to project.
4. Navigate to the first plots GeoDash widget in collection mode and inspect.

Desired outcome: The widget should display a numerical value for 'Elevation'. 'Elevation should not be 'N/A - N/A'. The widget should not time out.

### DIFF
--- a/src/py/gee/utils.py
+++ b/src/py/gee/utils.py
@@ -640,12 +640,20 @@ def getStatistics(extent):
     extentGeom = ee.Geometry.Polygon(extent)
     elev = ee.Image('USGS/GTOPO30')
     minmaxElev = elev.reduceRegion(
-        ee.Reducer.minMax(), extentGeom, 1000, maxPixels=500000000)
+        reducer=ee.Reducer.minMax(),
+        geometry=extentGeom,
+        scale=30,
+        bestEffort=True,
+        maxPixels=500000000)
     minElev = minmaxElev.get('elevation_min').getInfo()
     maxElev = minmaxElev.get('elevation_max').getInfo()
     ciesinPopGrid = ee.Image('CIESIN/GPWv4/population-count/2020')
     popDict = ciesinPopGrid.reduceRegion(
-        ee.Reducer.sum(), extentGeom, maxPixels=500000000)
+        reducer=ee.Reducer.sum(),
+        geometry=extentGeom,
+        scale=30,
+        bestEffort=True,
+        maxPixels=500000000)
     pop = popDict.get('population-count').getInfo()
     pop = int(pop)
     return {
--- a/src/py/gee/utils.py
+++ b/src/py/gee/utils.py
@@ -645,16 +645,16 @@ def getStatistics(extent):
         minmaxElev = elev.reduceRegion(**{
             'reducer':ee.Reducer.minMax(),
             'geometry' :extentGeom, 
-            'scale':0.1,
+            'scale':30,
             'bestEffort':True, 
             'maxPixels':1e13
             
         })
 
-        return {
+        return ee.Dictionary({
             'minElev': minmaxElev.get('elevation_min'),
             'maxElev': minmaxElev.get('elevation_max'),
-        }
+        })
     
     def reducePop ():
         popDict = ciesinPopGrid.reduceRegion(**{
@@ -664,7 +664,7 @@ def getStatistics(extent):
             'scale':927.67  
         })
         pop = ee.Number(popDict.get('population_count')).int()
-        return {'pop':pop}
+        return ee.Dictionary({'pop':pop})
 
     def sampleElve():
         centriod = extentGeom.centroid(1)
@@ -674,10 +674,10 @@ def getStatistics(extent):
             'maxPixels':1e13,
             'scale':30  
         })
-        return {
+        return ee.Dictionary({
             'minElev': sampleElv.get('elevation'),
             'maxElev': sampleElv.get('elevation'),
-        }
+        })
 
     def samplePop():
         centriod = extentGeom.centroid(1)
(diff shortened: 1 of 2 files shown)


Oliver-BE:
The first test is failing for me. I've followed the testing steps but I am getting the following results in my widget:
Screenshot from 2022-08-10 17-16-34

The second test works for me. This is what the widget looks like:
Screenshot from 2022-08-10 17-17-20

jdilger:
Hm well that's not right. Can you send me the GeoDash URL so I can recreate it on my end?

Oliver-BE:
I'm running this locally with a clone of the production database meaning that my URL likely won't work for you. Here are my project settings if you want to try to recreate this on your end:
Screenshot from 2022-08-10 17-45-59

jdilger:
I found a work around. And recreated the issue. I'll see what's going on.

jdilger:
@Oliver-BE ok I refactored the code. Can you retest when you when you have a moment? details on the changes below.

updates: The scale of the reductions for the population and elevation images needs to be different because of the difference in their resolutions. Typically we don't need to worry about this but for one reason or another the sum reducer is returning the wrong population (e.g. 0) when you have a small plot. Likewise when the plot size was smaller than the scale passed into the elevation min/max reducer it would return 'N/A'.  To circumvent this I am sampling the centroid of the plot when it's area is less than the spatial resolution of the population or elevation images.

Oliver-BE:
Now I'm seeing that the statistics widget for the project with a plot width of 1m is working fine:
Screenshot from 2022-08-12 16-01-54


But for the same project with a plot width of 10,000m I'm not getting any responses back on the same widget:
Screenshot from 2022-08-12 16-01-14

jdilger:
Hm ok I will need to take another look. How long did you wait to get a response? I think I was getting quick test times for a cached area. I tested it on a fresh aoi and after 2.4m it was able to complete. This is unusual though as a 10km extent should run almost instantly. 

image

jdilger:
@Oliver-BE ok this should be the final fix for this. I switched from the EE if statement to client side checking and it's passing quickly on my end.

src/py/gee/utils.py
```diff
@@ -639,17 +639,70 @@ def myImageMapper(img):
...
-    popDict = ciesinPopGrid.reduceRegion(
-        ee.Reducer.sum(), extentGeom, maxPixels=500000000)
-    pop = popDict.get('population-count').getInfo()
-    pop = int(pop)
-    return {
-        'minElev': minElev,
-        'maxElev': maxElev,
-        'pop': pop
-    }
+    ciesinPopGrid = ciesinPopGrid =ee.ImageCollection("CIESIN/GPWv411/GPW_Population_Count").sort('system:time_start',False).first()
```
Oliver-BE, line 642: I think `cieisinPopGrid` is a typo. Can you rename this to `citiesInPopGrid` (or `citiesinPopGrid`, I'm not sure the proper Python naming conventions)? 

Also, what is the `=ee.ImageCollection` syntax? I haven't seen that before.

Oliver-BE (commented):
The functionality is working for me now! I just have one small code comment.

src/py/gee/utils.py
jdilger: Oops definitely a typo with the double variable assigning. I had to look it up but I think it is supposed to be "ciesin" - it is an acronym for "Center for International Earth Science Information Network." But having that level of distinction is kind of pointless ha. I'll rename it to `population`.   

The `ee.ImageCollection` is one of [GEE's data structures](https://developers.google.com/earth-engine/guides/ic_creating). They're used to store raster's which you can filter, map, and reduce against. Similarly there is an `ee.FeatureCollection` which functions in the same way but for vector data.

Oliver-BE: Ah, gotcha - that makes sense. I thought perhaps there was some special GEE Python syntax that I hadn't come across yet since that didn't line didn't throw a (noticeable) error.

Thanks for the link to the GEE docs - GEE is something I don't have a lot of experience with but would love to learn more about.

Yeah I've normally seen Python code using `snake_case`, though it seems you're right in that GEE uses `camelCase` in their Python code to match their JavaScript API (`camelCase` is the standard in JavaScript).

jdilger: No problem on the documentation link - I am pretty much living in there haha. We have tons of GEE resources I can send your way if you want, or if you ever have a question I'm happy to chat or hack on it with you.

Oliver-BE: I'd love to take a look at any GEE resources you have, thank you! Feel free to DM/email them to me :student:

PR #1594, opened 2022-05-11
Fix projects CSV input/output processing

sifbiri:
## Purpose

The issue is we were splitting by commas and then zipmaping them with headers, so it all depends on the position of column and (what is after it). If last, anything after first comma will be lost (if a field has commas as part of its string value) .  We now check for the this case and handle it before going further. 

## Related Issues
Closes CEO-###

## Submission Checklist
- [X] Included Jira issue in the PR title (e.g. `CEO-### <title>`)
- [X] Code passes linter rules (`npm run eslint`/`clj-kondo --lint src`)

## Module(s) Impacted

Projects > Project Creation

## Testing

1. Create a new project.
2. When creating questions for the project, create a number of questions with commas in the question title and commas in the question answers.
3. Publish the project, and collect a few plots.
4. When collecting, make sure you answer some of the questions with the answers that include commas.
5. Go to the Project Management page and click the Download Plot Data and Download Sample Data buttons. 
6. Take a look at the downloaded CSVs.
7. All rows and commas of the downloaded CSVs should look normal, and there shouldn't be any unexpected rows and/or columns. See the attached image for an example.

Screenshot from 2022-06-27 12-35-44

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -1,0 +1,1 @@
+((nil . ((cider-clojure-cli-global-options . "-J--add-modules=jdk.incubator.foreign -J--enable-native-access=ALL-UNNAMED"))))
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,8 @@
 .settings/
 .tern-port
 collect-earth-online.iml
+.log
+.lsp
 
 # Build Directories and Files
 logs
--- a/deps.edn
+++ b/deps.edn
@@ -17,7 +17,8 @@
         ring/ring-json            {:mvn/version "0.5.0"}
         seancorfield/next.jdbc    {:mvn/version "1.1.613"}
         sig-gis/triangulum        {:git/url "https://github.com/sig-gis/triangulum"
-                                   :sha     "d2c5cab0985e0f660ff7ed1879f0fa4d0020003e"}}
+                                   :sha     "d2c5cab0985e0f660ff7ed1879f0fa4d0020003e"}
+        cider/cider-nrepl         {:mvn/version "0.28.3"}}
 
  :aliases {:build-db         {:main-opts ["-m" "triangulum.build-db"]}
            :config           {:main-opts ["-m" "triangulum.config"]}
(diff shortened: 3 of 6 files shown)


Oliver-BE:
I tried creating a project using the following configuration:

### Plot Design
Spatial Distribution: CSV file. I used the CSV file from the "Download example plot csv file" file.
Plot shape: Square
Plot width: 10
Assign Plots: No assignments
Quality control: none

### Sample Design
Spatial Distribution: Random
Number of samples: 10

I'm getting the following error:
Screenshot from 2022-06-27 11-50-33
Screenshot from 2022-06-27 11-50-38


`.dir-locals.el`

Why was this file added?

sifbiri:
`.dir-locals.el` is just a file containing project wide variables for emacs used with cider to launch REPL with arguments and connect to it.

Oliver-BE:
> `.dir-locals.el` is just a file containing project wide variables for emacs used with cider to launch REPL with arguments and connect to it.

It sounds like this should be added to the `.gitignore` and not to the repository directly

sifbiri:
I think it should be part of repo so that other users don't have to figure out the right JVM parameters to pass as we are using the python bridge to get the repl to launch in most common way for those users. Just like other *rc files to configure dev environment for the project.

Oliver-BE:
`src/clj/collect_earth_online/db/projects.clj`
Line 704: Can this be a private function? Also, can you add a short docstring to this function?
Line 710: Can you add a docstring to this function?

`src/clj/collect_earth_online/server.clj`
Lines 4, 5: Align `:as` and `:refer` columns with the rest of the required libraries.
Lines 82-83: What is the purpose of these changes?

sifbiri:
> Lines 82-83: What is the purpose of these changes?

To switch to use nREPL, a much richer REPL than default socket repl, developed by the community and provides server and client which allows editors to connect to it through common API/protocol (Like in Emacs using Cider package, VS Code using Calva extension, Sublime text using Sublimed, VIM using Fireplace...)  Helps a lot with debugging too. Will probably be part of Triangulum.

Oliver-BE:
Looks good except we normally have our `:import`s come before our `:require`s. In `src/clj/collect_earth_online/db/projects.clj` can you put the `:import` chunk before the `:require` chunk?

PR #1595, opened 2022-06-29
SER-4 SumofAnswers rules bug and showing saved answers

sifbiri:
## Purpose
We allow for the sum check rule of questions to be run correctly. And also we allow for the user to see the answer of different plots when collecting and moving about. Previously the answers would not show on UI in the input field. 

## Related Issues
Closes COM-###

## Submission Checklist
- [x] Included Jira issue in the PR title (e.g. `COM-### Did something here`)
- [x] Code passes linter rules (`npx eslint "src/js/**"`)
- [ ] No new reflection warnings (`clojure -M:check-reflection`)

#### Steps

1. Create a project with at least 2 questions with input of type number 
2. Add a sum rule for those questions (up to a number like 3) 
3. Start collecting on plots 
4. Input a number and then save it, move to the next plot 
5. Go back to the previous plot, the number inputed previously should still show up 
6. Go to next plot and save answer for it, such that its value combined with previous plot sums to bigger number set in the rule, hit save. 
7. You should see a popup telling you about the error, and giving you the constrained possible values that validate the rule for that input. 

## Screenshots
Values get shown when moving back to the plot. 

Screen Shot 2022-06-28 at 5 02 23 PM

### DIFF
--- a/src/js/collection.js
+++ b/src/js/collection.js
@@ -498,11 +498,9 @@ class Collection extends React.Component {
             mercator.samplesToVectorSource(visible),
             mercator.ceoMapStyles("geom", unansweredColor)
         );
-        mercator.enableSelection(
-            mapConfig,
-            "currentSamples",
-            sampleId => this.setState({selectedSampleId: sampleId})
-        );
+        mercator.enableSelection(mapConfig, "currentSamples", sampleId => {
+            sampleId != -1 && this.setState({selectedSampleId: sampleId});
+        });
     };
 
     featuresToDrawLayer = drawTool => {
@@ -796,7 +794,6 @@ class Collection extends React.Component {
 
                 mercator.highlightSampleGeometry(feature, color);
             });
-        this.setState({selectedSampleId: -1});
     };
 
     calcVisibleSamples = currentQuestionId => {
--- a/src/js/collection.js
+++ b/src/js/collection.js
@@ -498,9 +498,8 @@ class Collection extends React.Component {
             mercator.samplesToVectorSource(visible),
             mercator.ceoMapStyles("geom", unansweredColor)
         );
-        mercator.enableSelection(mapConfig, "currentSamples", sampleId => {
-            sampleId != -1 && this.setState({selectedSampleId: sampleId});
-        });
+        mercator.enableSelection(mapConfig, "currentSamples", sampleId =>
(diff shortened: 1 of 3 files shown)


Oliver-BE:
The saving functionality of the question answers is not working for me. Steps to replicate:

1. Create a project with two questions of type input number.
2. Add a sum rule for those two questions.
3. Go to collect a plot.
4. On the plot, select all samples and answer the first question with some number.
5. Click save.
6. Go to the second question and give an answer that would break the sum rule.
7. Click accept on the popup warning
8. Go back to the first question and try to change you answer and click save.
9. The new number won't be saved

sifbiri:
> The saving functionality of the question answers is not working for me. Steps to replicate:
> 
> 1. Create a project with two questions of type input number.
> 2. Add a sum rule for those two questions.
> 3. Go to collect a plot.
> 4. On the plot, select all samples and answer the first question with some number.
> 5. Click save.
> 6. Go to the second question and give an answer that would break the sum rule.
> 7. Click accept on the popup warning
> 8. Go back to the first question and try to change you answer and click save.
> 9. The new number won't be saved

I can't reproduce the steps. The number saves just fine after okaying the popup and returning to previous question.

sifbiri:
> Lines 82-83: What is the purpose of these changes?

To switch to use nREPL, a much richer REPL than default socket repl, developed by the community and provides server and client which allows editors to connect to it through common API/protocol (Like in Emacs using Cider package, VS Code using Calva extension, Sublime text using Sublimed, VIM using Fireplace...)  Helps a lot with debugging too. Will probably be part of Triangulum.

Oliver-BE:
I found another edge case here:

Steps to replicate:
1. Create a project with two input number questions.
2. Set the first to have a default value / placeholder value of 5.
3. Set the second to have a default value / placeholder value of 15.
4. Create a sum of answers rule where questions one and two must add up to 100.
5. Go to the collection page.
6. Select all samples, and click save on the first and second question without changing the default values (so the answer to question one is 5 and the answer to question two is 15). 

The test should fail because the answers add up to 20 (and not 100), but you are able to save this plot and continue.

There are also a number of other ways you're able to bypass the rule check. Perhaps (if we're not already) we should do this rule check when the user tries to save the entire plot vs save the answer to each individual question.

PR #1612, opened 2023-04-14
COL-326 clear missing user assignments

a-luz:
## Purpose

Remove user assignments for template projects from other institutions.

## Related Issues

Closes COL-326

## Submission Checklist

- [x] Included Jira issue in the PR title (e.g. `COL-### Did something here`)
- [x] Code passes linter rules for each file you updated. To lint all files at once, run `npm run lint`. To just lint one specific file run `npx quick-lint-js src/js/<file-you-changed> --snarky`.
- [x] No new reflection warnings (`clojure -M:check-reflection`)

### Module Impacted
Wizard > Template Loading

### Role

Admin

### Steps

1. Create a project from an institution;
2. Import a public template project from another institution;
3. Check that there are no users assigned to any plots;

### Desired Outcome

There should be no users assigned to plots in the Plot Design step of the project wizard.


src/js/project/CreateProjectWizard.js
```diff
@@ -199,11 +200,8 @@ export default class CreateProjectWizard extends React.Component {
     fetch("/get-project-imagery?projectId=" + projectId)
       .then((response) => (response.ok ? response.json() : Promise.reject(response)))
       .then((data) => {
-        const institutionImageryIds = this.context.institutionImagery.map((i) => i.id);
         this.context.setProjectDetails({
-          projectImageryList: data
-            .map((i) => i.id)
-            .filter((id) => institutionImageryIds.includes(id)),
```
sifbiri, line 206: I can't understand why we won't need this anymore?

a-luz: The sql function `select_imagery_by_project` already brings the imageries filtered by the institution_id, so we don't need to do the filtering in the front-end

PR #1678, opened 2023-10-10
Enhancement/clean nicfi dates

jonnvega:
## Purpose

Aurelie Shapiro reported that some dates' formatting for Planet NICFI imagery was not clear for users. Changed to use only one date. Working assumption is that the second date in the string is the correct one. Please see image below. 

## Testing

Planet NICFI must be visible in local workspace to test to locally. 

### Role

Admin

## Screenshots

image


jonnvega:
@a-luz any thoughts?

a-luz (approved):
LGTM!

PR #1687, opened 2023-10-19
Hide children cards if parent's answerId is not within their parentAnswerIds

jonnvega:
### Purpose

Questions cards were still visible even if their parent question's answer made them not applicable based on a set rule. 

### Testing

Created projects with parent questions in two levels and verified that it was reacting appropriately based on the rules I gave the project. 

### Role

Collectors.

### Desired Outcome

Questions are hidden if parent question's answer is not in their parent answers array:

https://github.com/openforis/collect-earth-online/assets/47796239/468c4097-0d03-4451-9f48-81ea36bc8d9a


a-luz (approved):
This enhancement worked just fine for the test cases I've made.

PR #1705, opened 2023-12-07
Hotfix/PlanetNICFIMenu not updating properly

jonnvega:
## Purpose

Fixing issue where the arrows actions were not making the correct requests. They were making the request for the previous layer. 

The state was updating properly but the dataLayer wasn't. 

## Testing

Please double check that the correct request is being made when either arrow is clicked. 

## Screenshots

https://github.com/openforis/collect-earth-online/assets/47796239/3a149c6a-cb0c-4aa0-942b-dc7f9143a776


a-luz:
Confirmed that it works:
nicfi-date

PR #1758, opened 2024-08-03
Return 500 on error for create project

jonnvega:
- Wrap call-sql "create_project" in a try/catch
- Return error and 500 if error in create_project


jonnvega:
If ok I'll proceed with the other handlers.

PR #1802, opened 2025-06-06
fixes data-response for registration and validation

jollyblondgiant:
## Purpose

we were having a couple issues relatd to registration: the email link that was being sent to verify did not have a "/" character between the base-url (eg "ceo-dev.sig-gis.com" or "local.collect.earth:8080") and the "verify-email" route. second, that the action of completing registration resulted in no data-response. This would be fine as the email sent is the prompt but the frontend expects a success code in order not to report a false error to the user.

## Related Issues

Closes COL-928, COL-957

## Submission Checklist

- [ x] Included Jira issue in the PR title (e.g. `COL-### Did something here`)
- [x ] Code passes linter rules for each file you updated. To lint all files at once, run `npm run lint`. To just lint one specific file run `npx quick-lint-js src/js/<file-you-changed> --snarky`.
- [ x] No new reflection warnings (`clojure -M:check-reflection`)

### Role

User

### Steps

1. From the login screen, click the "Register" buttton. Proceed to step 2.
2. If you have completed steps 4 and 5, go to step 9. otherwise, go to step 3.
3. enter an email address. If this is the email address of an existing user, proceed to  step 4. otherwise, skip to step 5.
4. Click the "re-send verification email" prompt. skip to step 6.
5. choose and confirm a password, agree to the TOS, and click "submit." go to step 6.
6. monitor the inboxes of the supplied email address. maybe check "all mail." go to step 7.
7.  Find the  email from "support".  follow the prompts in the email from step  6 to click the link to ceo. go to step 8.
8. confirm that verification completed successfully. Go to step 2.
9. you have completed testing.

### Desired Outcome

---

### DIFF
--- a/src/js/register.jsx
+++ b/src/js/register.jsx
@@ -19,13 +19,12 @@ class Register extends React.Component {
   fetchUserIfExists = () => {    
     fetch(`/check-email-taken?email=${this.state.email}`)
       .then((response) => Promise.all([response.ok, response.text()]))
-      .then(([response, data]) => {
-        console.log("email-exists-check response", {userId: data === "true"});
+      .then(([response, data]) => {        
 	this.setState ({userId: data === "true"});
       });    
   };
 
-  resendValidationEmail = () => {
+  resendValidationEmail = () => {  
     fetch (`/resend-validation-email?email=${this.state.email}`, {
       method: "POST"
     })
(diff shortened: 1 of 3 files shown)


src/js/register.jsx
```diff
@@ -57,6 +56,7 @@ class Register extends React.Component {
                                            alertMessage: "You have successfully created an account.  Please check your email for a link to verify your account."}}});           
             window.location = "/home";
           } else {
+            console.log(data);
```
a-luz, line 59: Remove console.log

src/js/verifyEmail.jsx
```diff
@@ -12,6 +12,7 @@ class VerifyEmail extends React.Component {
   }
   
   componentDidMount() {
+    console.log(this.props.email);
```
a-luz, line 15: remove console.log

